PR #9, opened 2015-01-02
Updates and revisions to nacl_amb and nacl_namd tutorials

KarlTDebiec:
I've updated the `nacl_amb` tutorial.

The major changes are as follows:
- Now compatible with Amber 14 (`ptraj` -> `cpptraj`)
- Cleaned up files and moved to better-organized subfolders
- Annotated Amber configurations
- Now use `sed` to set random seed to `$WEST_RAND16` rather than time; previously, segments started simultaneously from same parent would not diverge
- Eliminated `change_distance.py` and moved its function into `gen_istate.sh`
- Eliminated `ptraj.continue` and moved its function into `runseg.sh`
- Updated `coord_loader` to support NetCDF; `cpptraj` had unsolved issues reading coordinates from text
- Added `log_loader` to store time, energy, temperature, etc.
- Added `tar_segs.sh` to clean up `traj_segs` after running
- Added `cat_trajectory.py` to output binding or unbinding trajectories in `xyz` format


KarlTDebiec:
I've also updated the `nacl_namd` tutorial.

The major changes are as follows:
- Now use `sed` to set random seed to `$WEST_RAND16` rather than time; previously, segments started simultaneously from same parent would not diverge
- Added `log_loader` to store time, energy, temperature, etc.
- Added `post_iter.sh` to clean up `seg_logs` while running
- Added `tar_segs.sh` to clean up `traj_segs` after running

PR #10, opened 2015-01-18
Updated tutorials: replaced methane_gmx  with nacl_gmx; updated nacl_amb and nacl_namd

KarlTDebiec:
I've replaced the `methane_gmx` tutorial with an `nacl_gmx` tutorial in line with those written for Amber and NAMD. The methane/methane association tutorial had some issues:
- Lack of unbound state or periodic boundary conditions caused all probability to flow to outermost bin
- Random seed from time caused walkers branched simultaneously from the same parent to undergo identical dynamics

and it seemed appropriate to replace it with Na/Cl rather than maintain two different tutorials. `nacl_gmx` is consistent with `nacl_amb` and `nacl_namd`, they use the same WEST settings and store the same auxiliary data, though they use different force fields and therefore yield different results. `nacl_gmx` is configured to work with either 4.6 or 5.0 (or possibly earlier); I have tested it with 4.6.7 and 5.0.4.

I've also made the following corrections to the `nacl_amb` and `nacl_namd` tutorials:
- Bound target state corrected from second-closest bin to closest bin
- Store progress coordinate, coordinate, and log data for frame 0, rather than starting at frame 1
- Coordinate and log loaders cleaned up

I've written the corresponding updates to the three tutorials on the wiki, awaiting acceptance of this pull request. The text of the tutorials are identical except for the differences between the MD packages; and they can be maintained together using `vimdiff` or equivalent.


synapticarbors:
+1 for making the tutorials consistent between the various MD packages.

mczwier:
Very nice.

PR #11, opened 2015-03-11
First steps towards sphinx documentation

KarlTDebiec:
I have set up the basic framework for documenting the WESTPA project using sphinx.

I've ported over the Amber, Gromacs, NAMD and OpenMM NaCl tutorials from the wiki page, as well as the WWMGR page. The long-term plan is to [link this into] (http://blog.nikhilism.com/2012/08/automatic-github-pages-generation-from.html) an updated version of the [github-pages site] (http://westpa.github.io/westpa/). I've included the built html version of the documentation using Sphinx' default style; this is fine for now, but in the long term it would be nice for us to have a custom design; I plan to work on this sometime in the next month or two. @synapticarbors, you may want to check over the OpenMM tutorial and make sure that I haven't introduced any errors; I aligned the language to the other three NaCl tutorials in a few places but it remains mostly unique. I think in the long term it would nice for us to have this tutorial store the pcoord and coordinates identically to the other three (it may already, I have read but not worked through it), so that we only need to maintain one version of the analysis section of the tutorials, particularly as we add more information about how to analyze WESTPA simulations.

I haven't added any documentation of the actual code, but most functions/classes do not have detailed docstrings yet anyway. I suggest that we use [Google's docstring style] (http://sphinxcontrib-napoleon.readthedocs.org/en/latest/); this is much clearer than Sphinx's standard formatting, and is supported directly in Sphinx using the napoleon plugin (which will be included by default in the upcoming version, and is now in the beta). Converting those functions/classes that are documented already should not be too time-consuming.


ltchong:
Thanks for setting up the framework, Karl! Good idea to go with Google's docstring style.

synapticarbors:
@KarlTDebiec thanks, this looks awesome. 

One note though. I think that the [numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) is more common in the scientific community, and perhaps we should follow suite. I know that scikit-learn, pandas, IPython as well as msmbuilder and mdtraj all use it. 

The napolean plugin supports both:
http://sphinxcontrib-napoleon.readthedocs.org/en/latest/#google-vs-numpy

ltchong:
Good point, Josh. Since Google's docstring style is not really *that* different from the numpy docstring standard, we might as well go with a format that the current community of scientific programmers are used to seeing.

KarlTDebiec:
Yes, if all those projects use the numpy's format, I agree that it should be more familiar to our users than Google's and we should go with that.

mczwier:
This looks great, and I agree that numpy/scipy is probably the way to go.

mczwier:
My prior comment doesn't do this justice. This is very nice!

I think this is in great shape to merge unless anyone has immediate concerns.

ajoshpratt:
Looks brilliant to me.  Going forward, then, as we integrate new code with new classes and functions, should we be sure to include comments and documentation via NumPy style docstrings as a best practice?

mczwier:
That's the idea.

synapticarbors:
The following tool might be helpful for checking to see the state of the docs if we try to systematically go through and make sure we have good doc string coverage and it's conformant:

https://github.com/rmcgibbo/doccheck

It requires >python 3.3, but it is likely easy to backport it to 2.7, or just use anaconda to get the necessary version of python.

KarlTDebiec:
I've merged this, including the basic framework, the four tutorials, and the wwmgr page, into the master branch.

On my own fork, I have since pulled most of the remaining wiki pages through pandoc, cleaned them up, and integrated them into the sphinx documentation, including fixing most internal links between pages. I found a number of links to information on external websites; some targets such as bitbucket are probably fine, but others link to the Chong Lab redmine (behind a login) or to the replaced methane-gmx tutorial, and will need to be replaced or removed. Everything I saw looked like text suitable for inclusion into the main documentation on their own pages; I think it would be preferable to have all information included in this manner. I also noticed a few 'todo' links the pages' authors never got to.

Many of these pages haven't been updated for some time; about half of the currently-existing tools are documented (I've added stub pages for the other half) and I suspect a lot of the information is outdated. For this reason I don't think this content is ready to merge into the master branch yet. I currently have it in my fork in a [sphinx-wip branch](https://github.com/KarlTDebiec/westpa/tree/sphinx-wip). I suggest that we use a similar branch on this main fork to work on it until it is in a better state.

PR #13, opened 2015-03-31
Add warning to w_truncate to remove corresponding file-system data

synapticarbors:
This addresses the issue reported in westpa/west_tools#4 and gives explicit recommendations to the user to remove file-system data corresponding to truncated iterations.


mczwier:
Looks good!

PR #14, opened 2015-03-31
Force users to explicitly set WEST_ROOT in examples

synapticarbors:
On OSX `readlink` doesn't use `-f` flag for the same purpose as on Linux, so users attempting to use the `env.sh` script in all of the examples will get an error. 

I modified all of the `env.sh` scripts to bail if `WEST_ROOT` wasn't set, giving a message that the user should specify it. An alternative option would have been to use python as in the `west` shell script to make it platform agnostic. I chose not to take that path because I could imagine a user copying an example directory to a new location, and then the assumed position in the file hierarchy would not hold. The same is not true of the `west` script, which should always live in the same place.

### DIFF
--- a/lib/examples/nacl_amb/env.sh
+++ b/lib/examples/nacl_amb/env.sh
@@ -11,7 +11,11 @@ export CPPTRAJ=$(which cpptraj)
 
 # Inform WEST where to find Python and our other scripts where to find WEST
 export WEST_PYTHON=$(which python2.7)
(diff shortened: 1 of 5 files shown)


KarlTDebiec:
This is good; I never felt comfortable with the way they assumed they would be located in the examples directory.

synapticarbors:
@mczwier Any objections to merging this?

mczwier:
No objections.

PR #27, opened 2017-01-12
Install script

ajoshpratt:
A pretty simple script that should help with missing/confusing environment variables.

After install, a 'westpa.sh' shell script is created, which, when sourced or run with . , will set the necessary variables to run WESTPA.

It shouldn't conflict with any existing workflows, as it is entirely optional.


ajoshpratt:
I've tested this, but I do think if someone wants to test it in a weird situation, that would be helpful.

synapticarbors:
When I try to run this on OSX, I get the error:

```
sed: -i may not be used with stdin
sed: -i may not be used with stdin
sed: -i may not be used with stdin
```

From previous experience, `sed` has some differences between the version packaged with OSX and Linux. My recommendation would be to use a python script to do this so it's portable:

```python
import os

template = '''\
export WEST_ROOT={WESTROOT}
export WEST_BIN={WESTBIN}
export WEST_PYTHON={WESTPYTHON}

export PATH=$WEST_BIN:$PATH
'''

with open('westpa.sh', 'w') as f:
    f.write(template.format(WESTROOT=os.environ.get('PWD'),
                            WESTBIN=os.path.join(os.environ.get('PWD'), 'bin'),
                            WESTPYTHON=os.environ.get('WEST_PYTHON','')))
```

ajoshpratt:
Thanks for testing, Josh.  That's a great idea to use Python.  I'll make the changes.

Adam

synapticarbors:
No problem. I made a similar change a while back to get rid of `readline` in `/bin/west`: 3264caebb86446da6daa32cf47a295d589fbf55d

I think it's generally a good strategy to avoid linux tools that might not appear on every system or might behave differently.

ajoshpratt:
I agree!  I didn't realize sed had such differences.

I went ahead and made the changes, and testing in both an Arch VM and on OSX were successful.

The script also has the minor benefit of hardcoding the python it was compiled with, meaning that we shouldn't see weird errors attributable to that, if people use this.

Adam

PR #29, opened 2017-02-03
Added custom search bar

ASinanSaglam:
Added a search bar for searching westpa-user Google group. 

Please try breaking it and see if you can, I'm just passing the search terms to the URL and it should be safe but I can't guarantee it won't break. I tested it with various terms that might break it but I couldn't. 

You can test it out on my forked page: https://asinansaglam.github.io/westpa/

Also please let me know if you have any stylistic suggestions!


synapticarbors:
I would maybe say that the search bar should be down in the "Getting Help" section where the mailing list is mentioned rather than so large right at the top. Or possibly make the search bar smaller or less obtrusive.

ASinanSaglam:
I preferred to put it up top to make sure it's one of the first things the user sees and hopefully increase the usage of the mailing list archives to answer questions we have already answered in the mailing list. 

I do think it looked a bit clunky too, I made the bar a bit smaller (the text is slightly larger than the text in the page) and moved up the bar closer to the top (I tried doing that before but CSS is truly mysterious to me still).

ASinanSaglam:
After discussing with Lillian I made quite a few changes, removed the map, moved the bar to the help section and then moved the help section up above the funding section. At 100% zoom you still can't see it (well I can't on mine at least) but it's way closer to the top now.

ASinanSaglam:
Also matched the publications page format to the current main page.

ltchong:
Looks good!

PR #32, opened 2017-02-22
Sphinx documentation improvements

ASinanSaglam:
I have made some sweeping changes to the Alabaster theme we are currently using for our Sphinx documentation and did a bit more: https://asinansaglam.github.io/westpa/sphinx_index.html

- Content fonts changed to sans-serif (I probably should change the title fonts too)
- Contents now span the entire page, this helps particularly for some tutorials
- Sidebar color change, size change, added word-break 
- Added WESTPA logo, as a side note, I had to very slightly change the original so I get a background-less vector logo, had to fit it to the sidebar background
- Made some changes to conf.py for Sphinx to include the custom.css for each page generated
- Some README changes
- Fixed indentation levels for the RST files sphinx_index.rst and README.rst
- Converted [p53/Molecular-scale systems tutorial](https://chong.chem.pitt.edu/wewiki/Molecular-scale_systems) into [Sphinx](https://asinansaglam.github.io/westpa/tutorials/p53.html)

As a side note, I still have to request another pull request to the gh-pages branch. Can we somehow circumvent this by auto-generating Sphinx documentation every time someone makes changes to the docs folder or README.rst?

If yes, I am aware of some methods to do this but have never done it, any suggestions? I'm willing to try to implement it unless someone else wants to take it. Also note that I have already added a sub command to the make file called ghpages and docs can be updated just by doing

` make ghpages `

in docs folder (removing the leftover docs folder), commiting and pushing. 

Edit: As another side note, I'm aware that the website doesn't look great when viewed on mobile, I'll look into fixing that as I keep working on it. For some reason the % based sidebar widths don't play nice on mobile and some versions of browsers (Mac version has some differences, Chromium has some differences etc.).


KarlTDebiec:
Ali, here is what I use in my [sphinx makefile for Moldynplot](https://github.com/KarlTDebiec/MolDynPlot/blob/master/doc/Makefile):

```
gh-pages:
	make html
	git checkout gh-pages
	cd .. && \
    rm -vr _sources _static _images && \
    mv doc/_build/html/* . && \
    git add -A && \
    git commit -m "Updated documentation for `git log master -1 --pretty=short --abbrev-commit`" && \
    git push && \
    git checkout master
```

KarlTDebiec:
Also, I think that in order to automate the second push to gh-pages, you are looking for something like [this](http://stackoverflow.com/questions/16864259/how-to-run-bash-script-after-git-push). None of the solutions there are very elegant; ideally there would be a post-push hook that could be used to check if the second push should be done, but this doesn't appear to be part of git.

KarlTDebiec:
Sorry, clicked the wrong button.

ASinanSaglam:
Hi Karl, 

At the moment I have that part automated in a way, the command `make ghpages` is one I wrote and it does the search & replace + copying over the old one (I left out the cd ../ and rm commands, just in case I end up removing something by mistake, that is left to the user at the moment). It leaves out commit/push since I didn't think of testing if the authorization worked or it required a git configuration setup already. I should test it and add it to the current command I'm using. 

I was also thinking [github hooks](https://developer.github.com/webhooks/), they seem very very flexible but like you mentioned I also didn't see a post-push hook. I'm not sure if what I'm thinking can be done with it, maybe just a daily/hourly automated check (on the Chong lab server that hosts the wiki for example?) that keeps an updated repo, builds the documentation if the relevant files get changed?

Issue #24, opened 2016-09-19
nacl_amb example requires modification of env.sh to work


synapticarbors:
In the `env.sh` file in the nacl_amb example, there is no attempt to automatically detect where Amber is installed on a user's system:

https://github.com/westpa/westpa/blob/master/lib/examples/nacl_amb/env.sh#L5

The example therefore, does not work right out of the box, but there is no indication that the user needs to make any modifications before attempting to run the example. There also isn't any indication of this in the tutorial:

https://westpa.github.io/westpa/tutorials/amber.html

The example should be modified to attempt to automatically find the amber install and if it can't, print a useful error message.

ajoshpratt:
I think Amber always requires $AMBERHOME to be set, so we should probably make it use that instead of `which`.  Unless the local sysadmin has been behaving badly and not setting the variable within the environment module, which is possible.

But using $AMBERHOME/bin would allow them to use the amber.sh script that Amber suggests they use, so that's probably still the way to go.  We could set CPPTRAJ and PMEMD to $AMBERHOME/bin/cpptrajorpmemd, check if that exists, then just try `which` if it fails (IE, a last ditch effort, if they've just forced the Amber path into their $PATH variable).

PR #37, opened 2017-03-07
Fixed Amber Example

ajoshpratt:
Removed export of AMBERHOME line in env.sh in Na/Cl example.  Instead, it checks to see if the environment variable exists, and fails out with an error message if it doesn't.

While cpptraj/pmemd may exist on the path, if amberhome isn't set, it's unlikely that the libraries are in the LD_LIBRARY_PATH variable, and so it likely won't work.


ajoshpratt:
Referencing issue https://github.com/westpa/westpa/issues/24

ajoshpratt:
As there are no conflicts, this seems good to merge.  Merging.

PR #44, opened 2017-03-14
New Tool w_ipa

ajoshpratt:
Hello all,

This pull request is to integrate a new tool, w_ipa (WEST Interactive Python Analysis), a toolset designed for easy interactive analysis of WEST datasets.

You specify analysis schemes in your west configuration file (west.cfg), by default, and when the tool is run, it automatically loads/generates the analysis files as necessary.  It's designed for fast, simple analysis and to help springboard writing analysis or scripts (by handling file input/output), and to help with bookkeeping.

There are two new commands:
* w_ipa - launches you into an interactive ipython session on the command line where analysis is automatically performed, and datasets/properties from all WESTPA datasets are trivial to analyze.
* w_jupyter - helps with launching a jupyter/ipython notebook (mostly, it just ensures that all the WEST stuff is available on the PYTHONPATH such that it's easy to create a notebook)

The following file:
https://github.com/westpa/westpa/blob/wipython/lib/examples/odld/analysis.ipynb
is a jupyter/ipython notebook that demonstrates how to use w_ipa from the notebook, and some of the analysis it can perform.

The west.cfg has been updated to act as an example for this new interface and how to use it:
https://github.com/westpa/westpa/blob/wipython/lib/examples/odld/west.cfg

We should look into updating the other tutorial examples with an ipython notebook (and the west.cfg), too, once things are in place.  Crucially, there have been numerous changes to the Amber example, and so we'll need to reintegrate those to avoid conflicts.  To update those examples, we'll need to modify the west.cfg and generate an ipython notebook.


ajoshpratt:
TO DO:
Update copyright dates/info.

ajoshpratt:
To coincide with the new GROMACS tutorial which is coming later today, this merge request will be pushed by the end of the day.

PR #66, opened 2017-09-11
Fix the w.trace method from w_ipa

ajd98:
Due to an off-by-one error, the w.trace method from w_ipa previously traced segments from the previous iteration, rather than the current iteration.  This is fixed now.


ajoshpratt:
Thanks for catching this, @ajd98.

PR #69, opened 2017-10-01
Fixed routine for rate calculation from w_direct.

ajoshpratt:
Occasionally, the rate constant calculation from w_direct would return 0, even when there was probability in the istate AND a flux into the jstate.

This was caused by the rate normalization step for pairwise rate calculations (which is done by default); if no walkers had entered either the i or jstate, the sum of their probability was 0, which the istate probability was divided by, thus becoming 0/0, or NaN.  Once NaN exists in the dataset, any number interacting with NaN also becomes NaN; in that event, the code would return a 0 for the rate.

Now, if no probability has or does exist in the i or jstate, we don't do the pairwise rate normalization (and instead simply add 0 to the probability running average).  This problem primarily affected simulations where the bstates were different from the state definitions used in post processing, or with N>2 states where at least two were unpopulated for a period of time.

### DIFF
--- a/lib/west_tools/westpa/kinetics/_kinetics.c
+++ b/lib/west_tools/westpa/kinetics/_kinetics.c
@@ -895,7 +895,7 @@ struct __pyx_opt_args_6westpa_8kinetics_9_kinetics_sequence_macro_flux_to_rate {
   PyObject *stride;
 };
 
-/* "westpa/kinetics/_kinetics.pyx":602
+/* "westpa/kinetics/_kinetics.pyx":605
  * 
  * 
  * cdef class StreamingStats2D:             # <<<<<<<<<<<<<<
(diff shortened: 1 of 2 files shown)


ajoshpratt:
Given the impact (high) and number of changes (2 lines) of this merge request, if there no comments or concerns, I'll merge by 5 PM Eastern, October 2nd, 2017.

mczwier:
Good for me.

uadhikari:
Hi Adam,
I just used w_direct from the updated westpa version but in some cases it still gives 0.0 flux when there is some flux transferred between two states. It happens especially when I select some window of iterations.

ajoshpratt:
Hi Upendra,

That's concerning.  Do you think you can send me your h5 file and flags for direct.h5 via email?

Thanks for your patience on this.

Best,
Adam

PR #72, opened 2018-02-02
Removed pkill from initialization scripts, at request of @ajoshpratt

ajd98:
Adam requested that I remove the pkill commands from the init.sh scripts.  Testing with the odld_exe example suggests this is safe.

### DIFF
--- a/lib/examples/nacl_amb/init.sh
+++ b/lib/examples/nacl_amb/init.sh
@@ -11,11 +11,6 @@
 
 source env.sh
 
-# Make sure that WESTPA is not already running.  Running two instances of 
-# WESTPA on a single node/machine can cause problems.
-# The following line will kill any WESTPA process that is currently running.
-pkill -9 -f w_run
-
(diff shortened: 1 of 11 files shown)


mczwier:
pkill is a holdover from older, less-robust versions of the work manager. These changes should be fine. Before we merge, though, @ajd98, what work managers did you test with the odld_exe example?

ajd98:
@mczwier: I tested the `processes`, `threads`, and `serial` work managers. My tests consisted of running two simulations with a given work manager at the same time, and then checking that the simulations did not fail.  I also checked that the log files did not show strange similarities (by brief visual inspection). I'm not sure what issues used to arise when running more than one WESTPA simulation on a computer.

mczwier:
Sounds good. It wasn't about multiple simulations, it was about hung work managers and bad interactions with HDF5, IIRC. ZeroMQ tests with older versions of the work manager were particularly bad. It's doubtful that the pkill should ever have percolated out into the public, though. I'll go ahead and merge this.

PR #75, opened 2018-02-14
Added script for installing Miniconda followed by WESTPA

burntyellow:
This is the script for automatically installing Miniconda followed by WESTPA referenced in the documentation

https://github.com/westpa/westpa/wiki/Installing-WESTPA

### DIFF
--- a/miniconda_install_westpa.sh
+++ b/miniconda_install_westpa.sh
@@ -1,16 +1,16 @@
 #!/bin/bash
 
-if echo "$0" | grep '\.sh$' > /dev/null; then
-   echo 'Please source this script rather than executing it directly.'
-   echo "Usage: source miniconda_install_westpa.sh <Miniconda Installation Prefix>"
-   exit
+if ! echo "$0" | grep '\.sh$' > /dev/null; then
+   echo 'Please execute this script directly rather than sourcing the script.'
+   echo "Usage: ./miniconda_install_westpa.sh <Miniconda Installation Prefix>"
+   return 1
 fi
 
 if [[ "$#" -ne 1 || $1 == "-h" || $1 == "--help" ]]; then
    echo "Please specify the installation prefix for miniconda."
-   echo -e "Usage: . miniconda_install_westpa.sh <Miniconda Installation Prefix>"
+   echo -e "Usage: ./miniconda_install_westpa.sh <Miniconda Installation Prefix>"
    echo ""
-   return
+   exit
 fi
 
 PREFIX="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
@@ -18,7 +18,7 @@ PREFIX="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
 if [ -d $PREFIX ]; then
    echo "$PREFIX already exists."
    echo "Please specify a different directory or remove the current one and try again."
-   return
+   exit
 else
    mkdir -p $(dirname $PREFIX)
 fi
@@ -44,7 +44,7 @@ elif [ ${platform} == "Mac" ]; then
 else
    echo "Installation script only supports Linux or Mac OS X."
    echo "Exiting installation."
-   return
+   exit
 fi
 
 echo "download = " ${download}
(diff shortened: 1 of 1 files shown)


mczwier:
Glad to see the pull request for this.  The script works fine on Linux where wget is available, but fails on OS X where wget is not installed. Maybe the script can use curl instead of wget if wget is not available?

Also, this may have been part of the extended discussion about the westpa conda recipe, but it was unintuitive to me to go through the miniconda/westpa installation and then further have to execute the export and source westpa.sh commands. Is there a way to ensure that the conda installation puts links in the proper place, so that merely activating the conda environment is sufficient for putting w_run (etc) on the path?

burntyellow:
I've update the script to detect platform (Linux vs. Mac), use wget when available otherwise use curl, and automatically source westpa.sh when conda activate.  I've tested both on my Macbook pro and our cluster login node.

burntyellow:
Added verbose comments to script detail the delineation between miniconda installation and the westpa installation.  Also added a check for the number of commandline arguments and issue usage info if not satisfied.

mczwier:
Thanks, @burntyellow. I'll test and merge.

burntyellow:
Fixed the case where a user inputs a relative path for the PREFIX.  I'm done done now.

mczwier:
Kim,

This doesn't seem to run correctly if an install directory isn't specified. It fails to exit and blindly charges ahead. Do you need to `exit` instead of `return`? Does your testing protocol for this script cover all of the paths users might follow through it?

burntyellow:
The script was not meant to be executed directly but sourced.  I added logic in the beginning to check for this and to print an error message along with Usage syntax.  An exit statement instead of a return would log you out of the remote machine if the script was sourced, so "return" is retained.

I also made the "resolving absolute paths" command more portable by getting rid of readlink, which may not be available in OSX.  Should be a bit more clueless-proof now I hope.

burntyellow:
This script is meant to be executed on the commandline.  I'm updating the documentation as well.

synapticarbors:
Both `source activate` and `conda activate` work with newer versions of conda that support the latter, but it's probably better to stick with the new form of how you activate an environment (but maybe leave a note about the old syntax in case someone hasn't updated).

mczwier:
source activate fails for me, using the miniconda installed by the script.

burntyellow:
I tested the script under RHEL7 and OSX El Capitan, and both source activate and conda activate works.  What environment does this fail?

The required modification is not as simple as replacing source with conda.  I also need to configure the shell to use "conda activate" per the warning below but it's not a big deal.  I can make the changes and test more.

CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'.
If your shell is Bash or a Bourne variant, enable conda for the current user with

    $ echo ". /ihome/sam/kimwong/TEMP25/miniconda/etc/profile.d/conda.sh" >> ~/.bashrc

or, for all users, enable conda with

    $ sudo ln -s /ihome/sam/kimwong/TEMP25/miniconda/etc/profile.d/conda.sh /etc/profile.d/conda.sh

The options above will permanently enable the 'conda' command, but they do NOT
put conda's base (root) environment on PATH.

mczwier:
The environment that failed was my own Linux cluster. It could be (probably _is_) because I use zsh instead of bash, but I think the more important issue is the forward-compatibility that @synapticarbors brought up. If `conda activate` is the newer option, that's the one we should use. I think leaving it to the user whether to source `conda.sh` in `.bashrc` is wise; a y/n prompt would be best. The documentation can then reflect something like "Assuming conda is configured correctly, `conda activate westpa-2017.10` will make the WESTPA tools available at the command line."

burntyellow:
conda activate/deactivate is now the standard.  Please test soon and let me know if something comes up.  I've tested successfully on our Linux cluster and on my Mac.  Also, please take a read at the installation directions and feel free to directly make changes:

https://github.com/westpa/westpa/wiki/Installing-WESTPA

I followed @ synapticarbors's suggestion and made a note about the conda version syntax caveat.  Please change if you think there is a cleaner approach.

mczwier:
This looks good. Unless there are other issues or comments, I think it's time to merge.

PR #77, opened 2018-05-15
Adaptive voronoi binning plugin

ASinanSaglam:
I have written a simple plugin that implements the adaptive voronoi scheme tested in [Zhang 2010](https://aip.scitation.org/doi/full/10.1063/1.3306345), described in section III-D.  The code for the plugin is entirely based on the string method coded up by @synapticarbors and uses the [VoronoiBinMapper](https://github.com/westpa/westpa/blob/master/lib/west_tools/westpa/binning/assign.py) that's already in the code base. 

In short, an initial voronoi center is placed manually upon making the bin mapper in system and a new center is placed every n iteration (selected by user, defaults to n = 1, so one every iteration) and the the point that is the furthest away from its center among all the points in the last iteration is selected as the new center. This is repeated until a maximum number of centers is reached. 

Here is an example west.cfg plugin section to demonstrate the current options available: 
```
plugins:
  - plugin: westext.adaptvoronoi.AdaptiveVoronoiDriver
    av_enabled: true
    dfunc_method: odld_system.dfunc
    walk_count: 10
    max_centers: 100
    center_freq: 1
``` 
av_enabled can be used to enable/disable the plugin, dfunc_method is identical to the string method and is the function that the plugin will use to calculate distances between points, walk_count is the number of segments per center, max_centers is the maximum number of voronoi centers that will be placed and center_freq is the frequency of voronoi center placement in units of WE iterations.

I made a quick test simulation that is based on a 2 dimensional version of ODLD in [this repo](https://github.com/ASinanSaglam/WESTPA_2Dodld_adaptVor). You can clone [my fork of westpa](https://github.com/ASinanSaglam/westpa) and switch to adaptiveVoronoi branch, clone the test simulation repo under lib/examples folder after running the setup and it should just work. I also tested this extensively with a few different simulations that I'm working on currently. 

Please test the plugin if you have the time and let me know what features you would like to see and if you find any bugs or unexpected behavior (which I would love to hear in particular). I can definitely add more options to this but I would like to see what other people might use this for. I also thought it would be nice to have the plugin in place since the idea is already published in a paper. 

Edit: I purposefully didn't touch the copyright section at all since I wasn't sure what the exact procedure for it is, is there anything special I need to do or just update the date and name? 
Further edits: I will be merging the plugin within the week or early next week, it's needed for the upcoming WESTPA workshop.


ajd98:
I tested with the provided [example](https://github.com/ASinanSaglam/WESTPA_2Dodld_adaptVor) and Anaconda2.7-5.0.1, and it ran without error.  I appreciate that you included in the pull request explanations of the plugin's options; perhaps include these in a docstring as well?

ASinanSaglam:
Ah, documentation is indeed clearly lacking, will update soon with docstrings and include the information I mentioned here in the code. Thank you!

ASinanSaglam:
A few changes to this before merging it in:

1. I added more documentation in general. 
1. I wanted to allow for steady state simulations so I allowed a user defined function to make the next bin mapper. The function should take in the arguments the VoronoiBinMapper would and then return a bin mapper the user prefers. This allows the user to define a functional or a recursive bin mapper hence allowing you to embed the adaptive voronoi binning inside a particular region of space. I temporarily tweaked the ODLD example to show off this new feature (honestly I accidentally committed them, git lied to me and told me they wouldn't be committed). Note that I haven't tested this enough yet and I am not sure of the implications.

If you can give this a try at some point and let me know how it works out for you it would be great. I will test this as much as I can and then will revert the ODLD changes so that we can merge this into the main repo.

ASinanSaglam:
I have updated the copyright sections, added documentation regarding plugin options and tested the plugin further under steady state conditions. [Here](https://www.dropbox.com/s/vwlmrh9n2c11yfo/odld_vor_over.png?dl=0) is an example 2D ODLD result, the voronoi bins overlayed on top of the -ln of the probability. The walkers were recycled at the x = 10 and y =10 boundaries. [Here](https://www.dropbox.com/s/ny48j9u7w8u9sk9/odld_vor_over_2target.png?dl=0) is another test result with recycling at both 1.5 and 10.0 barriers in each dimension. The data visible outside of the barriers are due to the fact that these are recycled at the end of the iterations. The reflective boundary that exists in the original ODLD example is removed for these tests.

While I'm sure there are implications to running a steady state simulation using the embedded binning scheme I'm not thinking of, the plugin is currently running as intended and I think is ready to merge.

PR #80, opened 2018-07-05
Allow w_assign to pull the latest bin mapper and allow to pull bin mapper from a particular iteration

ASinanSaglam:
This small change brings w_assign on par with w_bins in pulling the latest bin mapper from the master h5 file. Currently w_assign will pull the first bin mapper created/bin mapper from the system.py from what I can tell from my tests. I additionally propose to enable --bins-from-h5file argument for w_assign and make it an integer argument instead of a store_true argument, allowing the user to select the iteration to pull the bin mapper from. **Sneaky edit:** Another option is to just add a secondary argument to select the particular iteration of course, which I'm starting to lean towards after playing with this. 

Pulling the latest mapper is necessary for supporting adaptive binning schemes and it's nice to allow the user to select a particular iteration to pull the mapper from as well. I'll test these changes to ensure it doesn't break anything but these should be decoupled from anything else from what I can tell. 

**Note:** I will clean up the comments, I just wanted to bring this up and see if anybody notices any issues first. 
**Further notes:** I will be trying to merge some form of this within the week since I think it's important to have this for the upcoming WESTPA workshop. At least pulling the latest bin mapper from the h5 file is key since otherwise it becomes difficult to show people how to use the bin mapper saved in the h5 file in w_assign (it's possible using bins-from-function but I'd rather avoid that entirely for the workshop).

### DIFF
--- a/lib/west_tools/w_assign.py
+++ b/lib/west_tools/w_assign.py
@@ -275,8 +275,12 @@ Command-line options
     def process_args(self, args):
         self.progress.process_args(args)
         self.data_reader.process_args(args)
+        # Necessary to open the file to get the current iteration
+        # if we want to use the mapper in the file
         self.data_reader.open(mode='r+')
-        self.n_iter = getattr(args,'bins_from_h5file',None) or self.data_reader.current_iteration
+        self.n_iter = self.data_reader.current_iteration
+        # If we decide to use this option for iteration selection: 
+        # getattr(args,'bins_from_h5file',None) or self.data_reader.current_iteration
 
         with self.data_reader:
             self.dssynth.h5filename = self.data_reader.we_h5filename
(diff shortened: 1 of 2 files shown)


ASinanSaglam:
I have removed the code that changes --bins-from-h5file option behavior, left only the change where we use the latest bin mapper from the h5 file by default. I used the same code path that w_bins uses for this so I don't think it's going to cause any differences in how people use w_assign (basically if you don't give binning options it just defaults to the bin mapper in the file) while making a major difference for people that use adaptive binning schemes and/or change binning during the simulation. I believe this is the better default behavior in general.

I will merge the code now for sake of the upcoming WESTPA workshop, if you spot a bug or would like me to change how this works we can open an issue and discuss it in detail.

PR #81, opened 2018-08-01
w_ipa fix

ajoshpratt:
w_ipa had some longstanding, minor bugs (and one I just found!) that prevented it from properly working in interactive mode.


ajoshpratt:
As I both require the code AND w_ipa is an isolated branch of the code, I feel comfortable merging.

PR #85, opened 2018-09-26
fixes #84

### LINKED ISSUE #84: WEST_CURRENT_SEG_ID seems broken
Hi all,

I keep seeing WEST_CURRENT_SEG_ID set to -1 in my runs. According to the [documentation](https://chong.chem.pitt.edu/wewiki/Environment_variables), this is not a valid value.

I think the problem lies in the line

https://github.com/westpa/westpa/blob/ba33dc3dd31de034eb23638ad234629a8aaca11f/src/west/propagators/executable.py#L322

which uses a too general if condition that eventually replaces every 0 by a -1.

```Python
>>> str(0 or -1)
'-1'
```
I can follow up with a PR.

Best,
Fabian

fabian-paul:


ASinanSaglam:
So, do we want to do string formatting a la ```"%d".format or "%d"%whatever.state_id``` like you suggested or we want to stick to str here?

mczwier:
.format() is Python 3 preferred

PR #88, opened 2018-10-23
Custom weights option for w_pdist

ASinanSaglam:
Hi all, 

This branch adds a new option called `--construct-wdataset` to w_pdist (I think `--construct-weights` is better) which allows you to pass in an array of weights instead of the weights stored in the westpa h5 file, achieved in the exact same style as `--construct-dataset`. I have tested this quite a few times and the branch works on a fresh docker container. The default behavior is unchanged where the weights are pulled from the westpa h5 file.

I have used this on various occasions now and I think it's useful enough to make it a general option. The simplest use case is to use a uniform distribution to get the walker distribution over your state space (to judge the sampling) but I have also used it to make probability distribution originating from a single parent (as in, a parent in the initial iteration, basically zero out every walker that didn't originate from that original parent and normalize).

I wanted to start this pull request to ask for feedback since I'm honestly not entirely certain if my changes to the data_reader are needed at all, I'm sure it could have been made a lot more elegant. So I'm completely open to suggestions to improve that one.


AnthonyBogetti:
This PR was ported to WESTPA 2.0 and will be released with WESTPA 2.0. Thank you Sinan for the feature. I am closing this PR for WESTPA 1.0.

PR #90, opened 2020-01-02
Typo: Corrected WM_PROCESS_ID to WM_PROCESS_INDEX

jdrusso:
WM_PROCESS_INDEX is an environment variable storing an identifier (modulo the number passed to --n-workers) for each spawned worker. It was mislabeled here as WM_PROCESS_ID


jeremyleung521:
Merged as part of #173.

PR #98, opened 2020-05-01
Minor bugfix: Changed check for seg_id defined

jdrusso:
The previous behavior checked `seg_ids == None`, which raises an error if `seg_ids` is defined. This was addressed by wrapping the whole thing in a try/except block.

However, the intended behavior can be produced by just checking with `is` instead of `==`, which correctly evaluates to `False` if `seg_ids` is defined instead of erroring out. Although the contents of the try block are fairly simple, doing the evaluation this way without the unnecessary try block prevents other potential errors in the block from being incorrectly ignored.

### DIFF
--- a/lib/west_tools/westtools/wipi.py
+++ b/lib/west_tools/westtools/wipi.py
@@ -325,11 +325,8 @@ class __get_data_for_iteration__(object):
         self.parent = parent
         current = {}
         current['iteration'] = value
-        try: 
-            if seg_ids == None:
-                seg_ids = range(0, iter_group['seg_index']['weight'].shape[0])
-        except:
-            pass
+        if seg_ids is None:
+            seg_ids = range(0, iter_group['seg_index']['weight'].shape[0])
         # Just make these easier to access.
         current['weights'] = iter_group['seg_index']['weight'][seg_ids]
         current['pcoord'] = iter_group['pcoord'][...][seg_ids, :, :]


synapticarbors:
My sense is that `==` should still produce a boolean no matter what the value of `seg_ids` is rather than raising an error, but using `is` for comparing to a singleton like `None` is the correct way to do things (see second bullet point of this section of PEP 8 https://www.python.org/dev/peps/pep-0008/#programming-recommendations).

jdrusso:
Indeed it does sort of return a boolean -- but since it's a numpy array, it's an array of booleans, where each element is checked for equality with `None`. The `if` check subsequently errors out, because now it's seeing an array of booleans instead of a single boolean.

synapticarbors:
@jdrusso, you're right. I didn't immediately see that `seg_ids` was a numpy array that has a different behavior than standard python objects.

jdrusso:
For what it's worth, I think this same behavior, and this same format of `if` statements checking invalid numpy arrays wrapped with a `try` block is present in a few other places in here, too. This was where I noticed it first though.

I think it came from the (now on the way out) Python 2.7 WESTPA code, where they weren't wrapped in the `try` blocks. I first came across this through the `westpa-2017.10` branch where I ran into it just throwing errors when `seg_ids` is valid.

mczwier:
This looks good to me. I agree with Josh that the right way to do this is with "is" rather than "==".

ltchong:
Thanks for the bugfix @jdrusso!

PR #117, opened 2020-08-07
Bugfix for states when auxref is None

jdrusso:
If no auxref is provided for a state (it's an optional parameter) then `state.auxref` will be set to `None`. To play nicely with the string formatting, this `None` object must be cast to a string `"None"`


synapticarbors:
@jdrusso -- just to be clear, is the code currently throwing a `TypeError: unsupported format string passed to NoneType.__format__` exception?

jdrusso:
@synapticarbors Yes, that's correct. Thanks for clarifying, I should've included that in the PR.

PR #118, opened 2020-08-10
Added functional tests for w_assign,direct and ipa.

atbogetti:
This also creates a ref/ directory where others can store reference h5 files for their tests.


synapticarbors:
Looks like this is failing due to a missing dependency,  `h5diff` and not passing the flake8 linting. To fix the dependency issue, you need to add `h5diff` to:

https://github.com/westpa/westpa/blob/westpa-2.0-restruct/devtools/conda-envs/test_env.yaml

assuming that it is available via conda. 

For the linting, you should locally install pre-commit, install as per the instructions and then fix the errors it outputs when you attempt to commit code to your local repo. Let me know if you run into issues.

synapticarbors:
oops, realized that `h5diff` is a local module, not an external dependency. Try:

`from .h5diff import H5Diff`

atbogetti:
I fixed the lint errors, could you take a look at the errors above?  I'm not sure why it cannot find my reference files.

synapticarbors:
@atbogetti --The temp directory you are making is relative to where the tests are being run, not the file that's being run. So on my machine (where I get the same error), if I `print(os.getcwd())` after the call to `os.chdir(temp.name)` in `test_w_assign.py`, I get:

```
/Users/xyz/Projects/github/westpa/tmpkav__2a7
```

which is why it can't find the `ref` directory. 

Here's an example change to the `w_assign` test that should work:

```python
import os
import shutil
import sys
import tempfile
import unittest

from .h5diff import H5Diff


sys.dont_write_bytecode = True


class Test_W_Assign(unittest.TestCase):

    test_name = 'W_ASSIGN'

    def test_run_w_assign(self):
        '''Testing if w_assign runs as expected and the assign.h5 file looks good.'''

        temp = tempfile.TemporaryDirectory(dir="./")
        os.chdir(temp.name)

        ref_dir = os.path.join(os.path.dirname(__file__), 'refs')

        shutil.copy2(os.path.join(ref_dir, 'west.cfg'), './')
        shutil.copy2(os.path.join(ref_dir, 'west.h5'), './')
        os.system("w_assign --config-from-file --scheme TEST")

        assert os.path.isfile('./ANALYSIS/TEST/assign.h5'), "The assign.h5 file was not generated."

        assign_ref_filename = os.path.join(ref_dir, 'assign_ref.h5')
        diff = H5Diff(assign_ref_filename, './ANALYSIS/TEST/assign.h5')
        diff.check()

        os.chdir("../")
        temp.cleanup()
```

A few other comments/questions:

- Why are you changing `sys.dont_write_bytecode`?
- If we can avoid changing directories, that might be desirable in case something fails we aren't left in a different directory

atbogetti:
Everything works now @synapticarbors, I removed the sys.dont_write_bytecode line and made sure it doesn't change directories.

synapticarbors:
Glad that worked. Also as a point of clarification, I don't think the tempdir was a bad idea -- I was more wondering if it was possible to just specify the `WEST_SIM_ROOT` to the temp directory so that you didn't have to change into that directory. This would be done in combination with something like `subprocess.run` with the `env` argument specified like (untested):

```python
test_env = os.environ.copy()
test_env['WEST_SIM_ROOT'] = temp_dir.name

...
subprocess.run('w_assign ...', env=test_env, ...)
```

PR #119, opened 2020-08-10
Update test_w_trace

poharrison:
Modified the w_trace_test file to include testing h5 file produced by the arg combos using h5diff.py. Added necessary reference files (in refs/ folder) and h5diff.py.


synapticarbors:
@poharrison -- I think this PR needs to be coordinated with the work that @atbogetti has in #118 since there are going to be conflicts between the two. It's not a big deal, but if #118 gets merged first, there is likely to be merge conflicts with your PR that will need to be resolved.

atbogetti:
I tried to resolve the merge conflicts but it led to a black formatting error in h5diff.py.  I'm not sure how to correct this and nothing in the file seems wrong.

synapticarbors:
When I ran `pre-commit run --all` on the branch, black inserts a line above `class H5Diff`. Do you have pre-commit set up?

atbogetti:
Okay I see it now and it has been fixed.

PR #120, opened 2020-08-10
Added a functional test for w_succ

darianyang:
Also added a w_succ output reference file (generated from ODLD west.h5 file) to the refs directory.


jeremyleung521:
Merged into #136.

PR #130, opened 2020-09-04
Incorporate Patrick Pisciuneri's update to the MPI work manager; add …

burntyellow:
…in dependency for mpi4py within setup.py. Tested on CRC HPC cluster.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,10 +115,12 @@ INSTALL_REQUIRES = [
     "matplotlib",
     "blessings",
     "ipykernel",
-    "mpi4py",
 ]
 
-EXTRAS_REQUIRE = {"tests": ["pytest", "pytest-cov", "nose"]}
+EXTRAS_REQUIRE = {
+    "tests": ["pytest", "pytest-cov", "nose"],
+    "mpi": ["mpi4py"],
+}
 
 EXTRAS_REQUIRE["dev"] = EXTRAS_REQUIRE["tests"] + ["pre-commit"]
 
--- a/setup.py
+++ b/setup.py
@@ -115,6 +115,7 @@ INSTALL_REQUIRES = [
     "matplotlib",
     "blessings",
     "ipykernel",
+    "mpi4py",
 ]
 
 EXTRAS_REQUIRE = {"tests": ["pytest", "pytest-cov", "nose"]}
--- a/src/westpa/work_managers/mpi.py
+++ b/src/westpa/work_managers/mpi.py
@@ -2,232 +2,356 @@
 A work manager which uses MPI to distribute tasks and collect results.
 """
 
+import sys, time, traceback
 import collections
 import logging
 import threading
-
+from collections import deque
 from mpi4py import MPI
-
 from westpa.work_managers import WorkManager, WMFuture
 
+log = logging.getLogger( __name__ )
 
-log = logging.getLogger(__name__)
-
-
+# +------+
+# | Task |
+# +------+
 class Task:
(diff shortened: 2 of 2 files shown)


synapticarbors:
Not sure if we need to do it in this PR, but it would be nice to figure out how to extend the tests in https://github.com/westpa/westpa/blob/westpa-2.0-restruct/tests/test_work_managers/test_mpi.py to do more than just check in the import.

synapticarbors:
Also, it looks like you didn't install pre-commit in your local repo. Please do that to satisfy the linter before committing:

https://pre-commit.com/#install

phpisciuneri:
Hi @burntyellow! - I am still here and alive - hope you all are well!  Let me know if I can help in some way.

synapticarbors:
It looks like mpi4py does not have non-windows wheels available and it is failing to build from source, which is causing the "Verify dev env" CI job to fail (since this is using pip instead of conda). 

I'm wondering if we should declare the mpi4py requirements as an "extra" (see https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies) so that users aren't pulling it in by default when using pip?

synapticarbors:
Also can we drop the "slave" naming convention and just call them "workers" or "nodes"? In the original implementation they were called "servers"/"clients".

burntyellow:
@phpisciuneri  Hey Patrick.  Good to hear from you.  The other MPI WM was deadlocking, I believe.  I tested your's extensively and it's legit ... so we're going with it.

synapticarbors:
@burntyellow -- thanks for making those changes. One thing about changing `is_master` to `is_manager` is that `is_master` is defined as part of the api used external to the work manager, so that should probably remain:

https://github.com/westpa/westpa/blob/2f418129d50e34def9947e0a8100278c5b2c1c6d/src/westpa/work_managers/core.py#L169-L173

and I'm guessing it needs to be available on all subclasses.

Also note that in python 3, `super()` does not need to specify arguments: 
https://www.python.org/dev/peps/pep-3135/

and it's probably best practice to just call it without arguments.

burntyellow:
> Not sure if we need to do it in this PR, but it would be nice to figure out how to extend the tests in https://github.com/westpa/westpa/blob/westpa-2.0-restruct/tests/test_work_managers/test_mpi.py to do more than just check in the import.

Agreed.  On the Todo list.

phpisciuneri:
> Also can we drop the "slave" naming convention and just call them "workers" or "nodes"? In the original implementation they were called "servers"/"clients".

Yes, please.  Thank you @burntyellow for making that change.

burntyellow:
@synapticarbors.  Thanks.  I'll check on is_master and is_manager.  The test that's running is on iteration 62 of 100.

phpisciuneri:
> The other MPI WM was deadlocking, I believe. I tested your's extensively and it's legit ... so we're going with it.

I can't remember why it didn't get merged in the first place.  I might have been having trouble with some of the example simulations (odld I believe) and didn't have the necessary domain expertise and things kind of petered out.  I apologize for that and would be really happy if this work is ultimately useful for you all.

burntyellow:
Josh, can you merge when you get a chance?  I successfully tested against the basic_nacl_amber_MPI-WM and odld_exe_MPI-WM examples; there's a pull request for these tests for Anthony.  There is is_master and "master" used in zeromq.  We should try to update the the language there as well but probably best after the MPI work manager is merged and tested by others.

PR #134, opened 2020-10-13
added additional tests for WESimManager

dnlebard:
Added a set of tests for WESimManager.

Still potentially need a test for WESimManager.initialize_simulation.


jeremyleung521:
Merged into #136.

PR #147, opened 2021-02-22
Fixes for w_ipa run_calculation and reading filenames from h5 file

SHZ66:
**Issue Number.** Is this pull request related to any outstanding issues? If so, list the issue number.  

**Describe the changes made.** A clear and concise description of what the problem is and what you did to fix it. E.g. [...] was happening and I've changed [...] to fix it.  

**Goals and Outstanding Issues.** A clear and concise list of goals (to be) accomplished.  
- [x] Fixed a bug where `bytes` is not properly converted into `str`when reading filenames for basis state files from west.h5.
- [x] FIxed a bug where bootstrapping length is inaccurately calculated causing `w_ipa` to crash. Note that `bootstrap_length` is only useful for logging purposes so this change would not lead to any actual result changes. 

**Major files changed.**  
- [x] Please see `file changed` section.

**Status.**
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the CONTRIBUTING document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

**Additional context.** Add any other context or screenshots about the pull request here.

PR #161, opened 2021-04-08
Checking binhash as bytes to eliminate errors (Fixes for WESTPA 1.0)

### LINKED ISSUE #142: Python3 bytes vs str causes duplicate bin mappers to be saved in the west.h5
It appears that when we made the switch from python 2 to 3, we didn't catch a difference in how the bin mapper hashes are represented. In the attributes of the iteration (e.g `data_manager.we_h5file['iterations']['iter_00000002'].attrs['binhash']`), this is a `str` (so unicode), but in the `bin_topologies` (e.g. `data_manager.we_h5file['bin_topologies']['index'][0][0]`) it is defined as `binhash_dtype = np.dtype('|S64')` which is `btyes` or more specifically `numpy.btypes_` but these are basically the same.

This leads to a few problems. In the `data_manager`, when we go to save the binning scheme for an iteration using `save_bin_mapper`, it attempts to find an existing duplicate of the bin mapper that's been used in a previous iteration using `self.find_bin_mapper(hashval)`. But `hashval` is a `str` but it's searching a collection of hashvals represented as `bytes`. Since `"abcd"` is not the same as `b"abcd"`, then `find_bin_mapper` raises a `KeyError` and the machinery thinks this is a new bin mapper even though it isn't. 

What we observe is that `bin_topologies` has an entry per-iteration, whereas it should only have a set of unique bin mappers. 

This could be fixed in a few different ways (and thanks for the suggestions from @dnlebard):
1. We convert the hashval to `bytes` in the function so that things just work whether you pass `str` or `bytes`.
2. We reconcile the storage so that the binhash is saved the same way throughout the whole system.

I'm sort of leaning toward (1) since that would allow for backward compatibility with simulations that have been run with this bug, although I'm interested in hearing competing views on this.

### LINKED ISSUE #146: UnicodeDecodeError raised when attempting to run w_assign
Hello, 

I was attempting to run w_assign on my own WESTPA data set but the following error was raised:

  File "xxx/miniconda3/envs/westpa/westpa-2020.03/lib/west_tools/westtools/binning.py", line 74, in mapper_from_hdf5
    mapper = pickle.loads(pkldat) 
UnicodeDecodeError: 'ascii' codec can't decode byte 0xbf in position 3: ordinal not in range(128)

I was able to recreate this error using the .h5 file and west.cfg from the [intermediate_p53 tutorial](https://github.com/westpa/westpa_tutorials/tree/main/intermediate_p53). Here are the steps I took to do so:

1. made a fresh conda install of latest westpa version (WEST version 19.9.0) into a new environment with python 3.7 
2. copied the unzipped p53.h5 and west.cfg from the tutorial into a clean directory
3. ran the following within that directory:  `w_assign -W p53.h5`
4. This gave the exact same error above

Not sure it matters, but this was all run on a linux system.
I believe this may be related to the most recent issue on this page (#142)

jeremyleung521:
#142 #146 #147 

Made fixes for binhash error where individual hashes in each iteration west['iterations']['iter_00000002'].attrs['binhash'] are saved as strings, while the ones in west['bin_topologies']['index'] are saved as byte strings. Whenever the hash is checked, it would check the byte string version of it, instead of the string version. The saving mechanism is unchanged, so it would continue to save string-type hashes per iteration. Should probably fix #142 and definitely fix #146. 

Also ported in #147 fixes as they're also related to byte strings.

- [x] Fix w_assign error where has is not found (#146)

**Major files changed.**  
- [x] lib/west_tools/westtools/binning.py 
- [x] src/west/data_manager.py 
- [x] lib/west_tools/westpa/h5io.py 
- [x] lib/west_tools/westtools/kinetics_tool.py 

**Status.**

NOTE: The resulting labels in the assign.h5 generated (state_labels, bin_labels) are still saved as byte strings.

### DIFF
--- a/lib/west_tools/westpa/h5io.py
+++ b/lib/west_tools/westpa/h5io.py
@@ -71,6 +71,19 @@ def calc_chunksize(shape, dtype, max_chunksize=262144):
     chunk_shape = tuple(chunk_shape)
     return chunk_shape
 
+
+def tostr(b):
+    '''Covert a nostandard string object ``b`` to str with the handling of the
+    case whhere ``b`` is bytes.'''
+
+    if b is None:
+        return None
+    elif isinstance(b, bytes):
+        return b.decode('utf-8')
+    else:
+        return str(b)
+
+
 #
 # Group and dataset manipulation functions
 #
--- a/lib/west_tools/westtools/binning.py
+++ b/lib/west_tools/westtools/binning.py
@@ -71,7 +71,7 @@ def mapper_from_hdf5(topol_group, hashval):
         for i in range(len(chunk)):
             if chunk[i]['hash'] == hashval:
                 pkldat = bytes(pickle_ds[istart+i,0:chunk[i]['pickle_len']].data)
-                mapper = pickle.loads(pkldat) 
+                mapper = pickle.loads(pkldat, encoding='latin1') 
                 log.debug('loaded {!r} from {!r}'.format(mapper, topol_group))
                 log.debug('hash value {!r}'.format(hashval))
                 return mapper, pkldat, hashval
--- a/lib/west_tools/westtools/binning.py
+++ b/lib/west_tools/westtools/binning.py
@@ -69,9 +69,9 @@ def mapper_from_hdf5(topol_group, hashval):
     for istart in range(0,n_entries,chunksize):
         chunk = index_ds[istart:min(istart+chunksize,n_entries)]
         for i in range(len(chunk)):
-            if chunk[i]['hash'] == hashval:
+            if chunk[i]['hash'] == bytes(hashval, 'utf-8'):
                 pkldat = bytes(pickle_ds[istart+i,0:chunk[i]['pickle_len']].data)
-                mapper = pickle.loads(pkldat, encoding='latin1') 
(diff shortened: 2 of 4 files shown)


jdrusso:
@jeremyleung521 Has this or an equivalent patch been added to westpa-2.0? 

I found the same error resulting from strings being read out of h5py as `bytes` instead of `str` occurring in a couple other places, which I fixed in aed6def1bfbad0f593820538b06af27c6d76fcf2 .

jeremyleung521:
> @jeremyleung521 Has this or an equivalent patch been added to westpa-2.0?
> 
> I found the same error resulting from strings being read out of h5py as `bytes` instead of `str` occurring in a couple other places, which I fixed in [aed6def](https://github.com/westpa/westpa/commit/aed6def1bfbad0f593820538b06af27c6d76fcf2) .

@jdrusso 
The auxref for sure: https://github.com/westpa/westpa/blob/westpa-2.0-restruct/src/westpa/core/data_manager.py#L537

Don't have the fix for the label or tstate though. I'll make a new PR with the similar to the h5io.tostr fix She made.

jdrusso:
Awesome, thank you very much! That h5io.tostr is great here too.

PR #214, opened 2021-12-01
haMSM Restarting Plugin

jdrusso:
**Issue Number.** N/A

**Describe the changes made.** 

Implemented restarting plugin for use with  https://github.com/westpa/msm_we.

When a WE simulation is run with this plugin, a number of simulations are actually run serially. Then, an haMSM is built from those, and all visited structures are weighted according to steady-state. A new set of simulations is restarted from these steady-state weighted structures. This process can be repeated.

To support this, I've added the notion of "start-states" to WESTPA. Start-states are passed to w_init  just like basis-states, and are used to initialize the simulation, but are NOT used for recycling. When I initiate a restart, the steady-state structures are added  as  start-states. Otherwise, they'd be used as recycling states, which would be putting the simulation in a quasistationary distribution instead of steady-state.

**Goals and Outstanding Issues.** 
- [x] Implement plugin for performing automated restarting using weighted structures from haMSM steady-state estimate
- [x] Implement "start states", which like basis states are used to seed initial walkers but unlike basis states are not recycled into
- [x] Documentation for plugin in API

**Major files changed.**  
- [x]  `w_init.py` Added arguments  to accept start-states
- [x] `w_run.py` Moved work-manager creation from the entry_point into  `run_simulation()`
- [x] `data_manager.py` Changed some `str` to `h5io.to_str`.
- [x] `core/propagators/executable.py` Added support for start-states, to correctly handle labeling segments initiated from start-states
- [x] `sim_manager.py` 
    - [x] Added  support for start  states. 
    - [x] Added a check  during plugin loading  to  see if  a plugin is  already loaded, and if  so,  to skip it. Details in the change about why this is necessary. 
    - [x] Modified probability output to show information about start-states (Note these probabilities don't necessarily add to 1, and some rescaling is typically necessary when using start-states)
- [x] `states.py` Added a new state  type for start  states. Also added a `basis_auxref` field, so start_states can point to the structure  file used to generate them.
- [x] `we_driver.py` Changed to  exclude start-states from the list of initial states.
- [x] `westext/hamsm_restarting/` Directory containing restarting  plugin and supporting  example files

**Status.**


jdrusso:
Note that 1567d37 addresses the issue mentioned in https://groups.google.com/u/1/g/westpa-users/c/hu8UL6Ig2e4/m/dnUSu8I9JQAJ

jeremyleung521 (approved):
Checked with an NaCl system and works flawlessly.  Approved to be merged.

PR #217, opened 2021-12-08
HDF5 Trajectory loader using the analysis API and a minor bug fix

SHZ66:

- [x] Added a Trajectory object for the trajectories generated by the HDF5 framework using the (newer) analysis APIs. 
- [x] A minor bug fix.

**Major files changed.**  
- [x] src/westpa/work_managers/core.py (Trajectory object for the HDF5 framework)
- [x] src/westpa/analysis/trajectories.py (`load_west` is not used or exposed to the users since we have the analysis modules now, but I fixed an issue with the pcoord because of the overlapping frame, anyway)
- [x] src/westpa/core/h5io.py (a minor bug fix where the traceback cannot be properly printed)

**Status.**


src/westpa/analysis/trajectories.py
```diff
@@ -255,6 +256,48 @@ def fget(walker, include_initpoint=True, atom_indices=None, sim_root='.'):
...
+class HDF5MDTrajectory(Trajectory):
+    """A MD trajectory stored by the HDF5 framework.
+
+    Parameters
+    ----------
+    atom_indices : list, default `None`
+
+    """
+
+    def __init__(self, atom_indices=None):
```
jpthompson17, line 268: Remove `atom_indices` parameter from constructor, and change `atom_indices=atom_indices` to `atom_indices=None` in definition of `fget()` function.

src/westpa/analysis/trajectories.py
```diff
@@ -255,6 +256,48 @@ def fget(walker, include_initpoint=True, atom_indices=None, sim_root='.'):
...
+        def fget(walker, include_initpoint=False, atom_indices=atom_indices):
+            from westpa.core.h5io import WESTIterationFile
```
jpthompson17, line 270: Move import to top of file per PEP 8.

src/westpa/analysis/trajectories.py
```diff
@@ -255,6 +256,48 @@ def fget(walker, include_initpoint=True, atom_indices=None, sim_root='.'):
...
+
+            iter = walker.iteration
```
jpthompson17, line 272: Could we spell out "iteration"? (This follows the style in the README and avoids shadowing the built-in `iter`.)

src/westpa/analysis/trajectories.py
```diff
@@ -255,6 +256,48 @@ def fget(walker, include_initpoint=True, atom_indices=None, sim_root='.'):
...
+            if 'trajectories' not in iter.h5group:
+                raise KeyError("The loaded west.h5 does not seem to be generated by the HDF5 framework")
```
jpthompson17, line 274: From the user's perspective, does it make sense for this to be a `KeyError`? I'm wondering if maybe it should be a `ValueError`.

src/westpa/analysis/trajectories.py
```diff
@@ -255,6 +256,48 @@ def fget(walker, include_initpoint=True, atom_indices=None, sim_root='.'):
...
+            with WESTIterationFile(traj_filename) as traj_file:
+                traj = traj_file.read_as_traj(iteration=iter.number, segment=walker.index, atom_indices=atom_indices)
+
+            if include_initpoint:
+                parent = walker.parent
+                if isinstance(parent, InitialState):
+                    bstate = parent.basis_state
+                    iter0 = Iteration(0, iter.run)
+                    parent = iter0.walker(bstate.state_id)
+                else:  # shouldn't happen but handles the case anyway
```
jpthompson17, line 287: What about the case that the parent is a `Walker`?

jpthompson17 (commented):
I think the logic in `HDF5MDTrajectory` still needs a bit of work. For example, it's perfectly legal to do
```
trajectory = HDF5MDTrajectory()
traj = trajectory(walker, include_initpoint=True)
```
even when `type(walker.parent) is Walker`.

SHZ66:
> 

Yes, I accidentally included the case in the `else` block. This is fixed.

src/westpa/analysis/trajectories.py
SHZ66: It is imported here to avoid the whole analysis module being dependent on `h5io`.

jpthompson17: I wouldn't worry about it--it's already done here: https://github.com/jpthompson17/westpa/blob/d738f18eb030482fcd0e5832a0b301a88baa669f/src/westpa/analysis/core.py#L7

jpthompson17 (approved):
Looks good. Thanks, She!

jeremyleung521:
A couple things (that might be unrelated to the scope of this PR):
- I didn't catch this before, but it seems like Run excludes the last iteration by default. This will be difficult for people who `w_truncate`-ed to the desired length but can't access the last iteration.
- It won't load any trajectory with a `start_state` because it would hit an out of range error [here](https://github.com/westpa/westpa/pull/217/files#diff-fd81ae7e5f3026d581a5670d5faafefec44dcc35f5e317e0eee61e5a20f8e8beR308) since `basis_state_id > len of group(basis_state_id)`. Might be worth throwing an exception for `InitialState.ISTATE_TYPE_START` ([link](https://github.com/westpa/westpa/blob/westpa-2.0-restruct/src/westpa/core/states.py#L154)) now. It should be fixed once we have the fallback code to read from the bstates data ref files instead. 

Both can be checked with the west.h5 file in https://github.com/westpa/westpa2_tutorials/tree/main/basic_nacl/w_crawl
- iter 2 is not accessible.
- iter 1 walkers 15-19 are loadable, the rest can't because of the sstate issue.

jpthompson17:
@jeremyleung521 Thanks for catching these issues. I do think fixing them is outside scope of this PR, since it will involve changes to `westpa.analysis.core`. I've created a new issue #225, which I will address in a separate PR.

jeremyleung521 (approved):
Ready to merge after the docs are added (PR #31 in She's fork).

SHZ66:
> Ready to merge after the docs are added ([PR #31 in She's fork](https://github.com/SHZ66/westpa/pull/31)).

Thanks for the PR, Jeremy! I just merged it.

PR #221, opened 2022-01-05
Added in the option to specify min/max weight thresholds in west.cfg (Attempt #2)

atbogetti:
Added in support for minimum and maximum weight thresholds for WESTPA 2.0

Describe the changes made. Oftentimes during a weighted ensemble simulation, trajectories reach very low weights through many rounds of splitting. Conversely, many rounds of merging in the initial bins may lead to massively-weighted trajectories. This change gives users the ability to set minimum and maximum weight thresholds for their simulation where if a walker's weight falls below or above the given thresholds it will not be split or merged during that round of resampling. To do this, I added the relevant logic into we_driver.py which, after splitting or merging, will check the new weights up against the thresholds.  If the weights cross the thresholds, the splitting or merging if undone. I also modified the while loops so that they don't become infinite in the case of no splitting or merging that does not help to achieve the target number of walkers (this is in the adjust_counts function). Because of the way this was implemented, the number of walkers per bin will not be exact as if there are 4 walkers in a bin and the target count is 5 and none of those four walkers are eligible for splitting, that bin will have one fewer than the target. Since the target number being reached is not essential for the WE algorithm, this is fine. Note that this implementation is now compatible with the new grouping (bin-less) resampler.

To enable this, add the following to the `we` section of your `west.cfg` file:

    smallest_allowed_weight: 0.01
    largest_allowed_weight: 0.25

Also, for checking the results of simulations in terms of summed weights and what the actual min/max weights were per iteration please feel free to use my testing script [here](https://github.com/atbogetti/odld/blob/main/analysis/get_probs.py).

- [ ] Testing and Documentation

**Major files changed.**  
- [ ] we_driver.py

**Status.**


src/westpa/core/we_driver.py
```diff
@@ -559,19 +594,33 @@ def _adjust_count(self, bin, groups, target_count):
...
                 # KFW CHECK BACK                    try:
                 # KFW CHECK BACK                        new_segment.id_hist = list(segments[-1].id_hist)
                 # KFW CHECK BACK                    except AttributeError:
                 # KFW CHECK BACK                        pass
-                i.update(new_segments_list)
-                bin.update(new_segments_list)
 
                 if len(bin) == target_count:
                     break
+                elif i == sorted_groups[-1]:
```
SHZ66, line 617: The unit test is failing because of this line. The original code will run a few iterations of the `while` loop (and therefore, several times of the for-loop inside) to adjust the count to the target value, whereas this line will set the target count to `len(bin)` and break too early.

src/westpa/core/we_driver.py
```diff
@@ -588,13 +637,20 @@ def _adjust_count(self, bin, groups, target_count):
...
+                    else:
+                        # print("no threshold break (bw)")
+                        i.add(merged_segment)
+                        bin.add(merged_segment)
                     # As long as we're changing the merge_walkers and split_walkers, adjust them so that they don't update the bin within the function
                     # and instead update the bin here.  Assuming nothing else relies on those.  Make sure with grin.
                     # in bash, "find . -name \*.py | xargs fgrep -n '_merge_walkers'"
                     if len(bin) == target_count:
                         break
+                    elif i == sorted_groups[-1]:
```
SHZ66, line 652: Same as above.

SHZ66 (changes requested):

jeremyleung521 (approved):
With the additional checks I just pushed, I approve of the changes.

PR #220, opened 2022-01-03
Making sim_manager.initialize() arguments explicit in w_init

jeremyleung521:
**Issue Number.** Is this pull request related to any outstanding issues? If so, list the issue number.  

**Describe the changes made.** A clear and concise description of what the problem is and what you did to fix it. E.g. [...] was happening and I've changed [...] to fix it.  
When using old custom sim managers where `start_states` are not defined, the following error message (which is not descriptive) might appear. I have made the keyword arguments explicit in lines 216-217 for w_init.py so the Type Error will claim `start_states` is an unexpected keyword argument, a much more descriptive error message. 

Previously w/o fix: 
> System is being built only off of the system driver
> Traceback (most recent call last):
>   File "/user/conda_local/envs/westpa-dev/bin/w_init", line 33, in <module>
>     sys.exit(load_entry_point('westpa', 'console_scripts', 'w_init')())
>   File "/user/conda_local/westpa-2.0dev1/src/westpa/cli/core/w_init.py", line 118, in entry_point
>     args.shotgun,
>   File "/user/conda_local/westpa-2.0dev1/src/westpa/cli/core/w_init.py", line 216, in initialize
>     basis_states, target_states, start_states, segs_per_state=segs_per_state, suppress_we=shotgun
> TypeError: initialize_simulation() got multiple values for argument 'segs_per_state'

Now w/ fix: 
> System is being built only off of the system driver
> Traceback (most recent call last):
>   File "/user/conda_local/envs/westpa-dev/bin/w_init", line 33, in <module>
>     sys.exit(load_entry_point('westpa', 'console_scripts', 'w_init')())
>   File "/user/conda_local/westpa-2.0dev1/src/westpa/cli/core/w_init.py", line 110, in entry_point
>     initialize(
>   File "/user/conda_local/westpa-2.0dev1/src/westpa/cli/core/w_init.py", line 217, in initialize
>     sim_manager.initialize_simulation(
> TypeError: initialize_simulation() got an unexpected keyword argument 'start_states'

**Goals and Outstanding Issues.** A clear and concise list of goals (to be) accomplished.  
- [x] Better error message when using start_states with a custom sim_manager where start_states are not defined.

**Major files changed.**  
- [x] /src/westpa/cli/core/w_init.py

**Status.**
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] I have read the CONTRIBUTING document.
- [x] My code follows the code style of this project.
- [x] All new and existing tests passed.

[...]

PR #222, opened 2022-01-06
Making Start States Compatible with the HDF5 Framework

jeremyleung521:

Start states do not work with the HDF5 Framework. This PR fixes that by allowing the propagator to read the start states structure files directly (since they are not saved into any of the per-iteration HDF5 files). 

As a bonus, I expanded the docstrings in src/westpa/core/states.py and fixed some typo errors here and there.

Should be merged after #220 since it contains those changes. (Or just close #220 and merge this one directly)

- [x] Make start states compatible with HDF5 Framework

**Major files changed.**  
- [x] src/westpa/core/data_manager.py
- [x] src/westpa/core/sim_manager.py

**Status.**

### DIFF
--- a/src/westpa/core/data_manager.py
+++ b/src/westpa/core/data_manager.py
@@ -1113,7 +1113,7 @@ class WESTDataManager:
 
         if not self.store_h5:
             return
-        print('in prepare_segment_restarts')
+
         for segment in segments:
             if segment.parent_id < 0:
                 if initial_states is None or basis_states is None:
--- a/src/westpa/core/data_manager.py
+++ b/src/westpa/core/data_manager.py
@@ -1113,13 +1113,21 @@ class WESTDataManager:
 
         if not self.store_h5:
             return
-
+        print('in prepare_segment_restarts')
         for segment in segments:
             if segment.parent_id < 0:
                 if initial_states is None or basis_states is None:
                     raise ValueError('initial and basis states required for preparing the segments')
                 initial_state = initial_states[segment.initial_state_id]
-                basis_state = basis_states[initial_state.basis_state_id]
+                # Check if it's a start state
+                if initial_state.istate_type == InitialState.ISTATE_TYPE_START:
+                    log.debug(
+                        f'Skip reading start state file from per-iteration HDF5 file for initial state {segment.initial_state_id}'
+                    )
+                    continue
+                else:
+                    basis_state = basis_states[initial_state.basis_state_id]
+
                 parent = Segment(n_iter=0, seg_id=basis_state.state_id)
             else:
                 parent = Segment(n_iter=segment.n_iter - 1, seg_id=segment.parent_id)
--- a/src/westpa/core/propagators/executable.py
+++ b/src/westpa/core/propagators/executable.py
@@ -92,23 +92,27 @@ def restart_loader(fieldname, restart_folder, segment, single_point):
 
 
 def restart_writer(path, segment):
-    '''Prepare the necessary files from the per-iteration HDF5 file to run ``segnment``.'''
+    '''Prepare the necessary files from the per-iteration HDF5 file to run ``segment``.'''
     try:
         restart = segment.data.pop('iterh5/restart', None)
         # Making an exception for start states in iteration 1
-        if segment.n_iter != 1:
-            if restart is None:
-                raise ValueError('restart data is not present')
+        if restart is None:
+            raise ValueError('restart data is not present')
 
         d = BytesIO(restart[:-1])  # remove tail protection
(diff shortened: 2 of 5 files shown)


src/westpa/core/propagators/executable.py
```diff
@@ -95,15 +95,20 @@ def restart_writer(path, segment):
...
         with tarfile.open(fileobj=d, mode='r:gz') as t:
             t.extractall(path=path)
-
     except Exception as e:
-        log.warning('could not write restart data for {}: {}'.format(str(segment), str(e)))
+        if segment.n_iter != 1:
+            log.warning('could not write restart data for {}: {}'.format(str(segment), str(e)))
+        else:
+            d = BytesIO()
+            log.warning('In iteration 1, so assuming you\'re trying not to read a start state from per-iteration HDF5 file.')
```
jdrusso, line 111: Here, why don't we specifically catch the ValueError you raise above (or we can raise a custom exception). It's totally plausible other unrelated stuff could break in iteration 1, so let's be parsimonious and minimal about what exceptions we try to handle

src/westpa/core/sim_manager.py
```diff
@@ -286,8 +286,18 @@ def initialize_simulation(self, basis_states, target_states, start_states, segs_
...
-        self.get_bstate_pcoords(start_states)
-        self.report_basis_states(start_states)
+
+        try:
+            if start_states[0].state_id is None:
+                last_id = basis_states[-1].state_id
+                for start_state in start_states:
+                    start_state.state_id = last_id + 1
+                    last_id += 1
+        except IndexError:
```
jdrusso, line 296: It looks like this IndexError would only ever be raised if the start states are empty, yeah? Or if basis_states is of length 0. If that's accurate, let's just handle that with the logic (i.e. `if .... and len(start_states) > 0:`)

src/westpa/core/data_manager.py
```diff
@@ -1119,7 +1119,15 @@ def prepare_segment_restarts(self, segments, basis_states=None, initial_states=N
                 if initial_states is None or basis_states is None:
                     raise ValueError('initial and basis states required for preparing the segments')
                 initial_state = initial_states[segment.initial_state_id]
-                basis_state = basis_states[initial_state.basis_state_id]
+                # Check if it's a start state
```
jdrusso, line 1122: I think we talked about this logic, but could you add a little comment here describing why it's OK to skip adding this data? Or what'll be missing from the per-iteration H5 file? From our discussions, I think the rationale was it's fine to omit this from the h5, because even if this run is completed and started later using the HDF5 framework, the basis states will be used again but the start states will not. So, not storing start states is fine for that reason. If I recall correctly, let's just clearly document that here.

I think you explained to me at one point, but it looks like this is going to leave some segment data in the per-iteration H5 empty. I think this is what you mentioned might cause issues with `westpa.analysis` -- it would be great if you can clearly document why/what specifically those issues might be here. (If this causes problems down the line, or if we decide the above behavior is not acceptable after all, it'd be great to have that written down here so we don't have to rediscover it then)

jdrusso (changes requested):
Looks great! Just a few notes on minor things. Thanks a lot for doing this, this is super helpful.

src/westpa/core/propagators/executable.py
jeremyleung521: The big exception block, I can only assume (as it's written by @SHZ66) is there to catch other FileExistsError and such from the BytesIO/tarball as well. I broke the ValueError to a specific except block, hopefully making it more clear. Maybe She can comment.

src/westpa/core/data_manager.py
jeremyleung521: Added in a big block of comments! Thanks!

src/westpa/core/sim_manager.py
jdrusso: If you do it as `if len(start_states) > 0 and start_states[0].state_id is None:` (i.e. put the len() check first), it'll short circuit if that first check isn't passed, and will only attempt the second if the length is greater than zero

jeremyleung521: Thanks for the tip! It works and is fixed!

jdrusso (approved):
Thanks! Looks great now.

jdrusso:
Just coming back to this -- I'm trying to do some more sophisticated stuff with combining the H5 framework with start-states, and this solution no longer works well. 

If there's a genuine error in loading restart data which should exist (but doesn't) this falsely just assumes it's a start-state and carries on. We need some logic to more explicitly check whether we expect to get data back in the first place.

PR #223, opened 2022-01-11
Added option for constant number of total walkers

atbogetti:
This PR adds about 50 lines of code to `we_driver.py` which allow a user to keep the total number of walkers constant throughout a WESTPA simulation. There were a few different ways I think that this could be implemented and I tried to avoid a global comparison step since the current driver resamples by bin.  I was able to keep resampling by bin while still comparing to a global total number by dividing up that global total and re-assigning bin target counts before resampling occurs.  If a user requests 100 total walkers then the bin target counts are reassigned to be 0 if that bin is currently unoccupied and 100/(number of occupied bins) for the rest. This allows for essentially no change in the driver's main algorithm.  I also added code to handle cases if the total walkers divided by the number of occupied bins does not come out to an even number.  In this case, extra walkers are merged (subtracted from the bin target count) in the most occupied bin and needed walkers are split (added to the bin target count) in the least occupied bin.  This has worked well for me in functional tests with ODLD.  Users can add the following to the `we` section of `west.cfg` to turn on this feature:

     we:
       constant_walkers: true
       total_walkers: 50

- [ ] This is currently not compatible with the thresholds PR, so I need to figure out a way to adjust the thresholds code accordingly
- [ ] Documentation
- [ ] Testing

**Major files changed.**  
- [ ] `we_driver.py`

**Status.**


atbogetti:
Before this gets merged I need to account for cases where bin occupancy may be greater than the total walkers in the simulation (for instance, if 100 bins are occupied but there are only 50 total walkers in the simulation).  I think that one way to handle this is to check and make sure the number per bin never reaches zero (probably unlikely but that would be a good test) but also to encourage the user in that case to increase the number of total walkers.  This may be more of a problem with MAB since the bin occupancy is so high but could be in other cases as well.

AnthonyBogetti:
I'm going to close this for now.  We will figure out a way to add in the weights to the target counts and then re-open.

PR #240, opened 2022-05-11
A new BinlessMapper for WESTPA

atbogetti:
This PR adds a new BinlessMapper to WESTPA that allows for the complete employment of binless schemes during a weighted ensemble simulation.  This is essentially just a special case of the FuncBinMapper that takes in a user-defined function that groups progress coordinate values.  The framework here for use is very similar to the MAB scheme in that it can be placed inside of a Recursive Bin for use with a target state.  Please see my ODLD binless or binless-2D branch for an example of how to get this ready in the west.cfg file.  I also re-named the current group-related functions to be subgroups since those technically operate within existing "groups". See documentation in the Wiki's Binning page for how to use.

**Goals and Outstanding Issues.**
- [ ] Make this usable with more than 2 dimensions
- [ ] Remove requirement for needing to know the total number of groups before-hand

**Major files changed.**  
- [ ] we_driver.py (just for the subgroup renaming)

**Status.**

### DIFF
--- a/src/westpa/core/_rc.py
+++ b/src/westpa/core/_rc.py
@@ -305,6 +305,13 @@ class WESTRC:
         elif nestmapname == 'MABBinMapper':
             sim_manager = westpa.core.binning.mab_manager.MABSimManager(rc=self)
 
+        elif binmapname == 'BinlessMapper':
+            sim_manager = westpa.core.binning.binless_manager.BinlessManager(rc=self)
+            sys.stderr.write("-- WARNING  [config] -- Do not use BinlessMapper as an outer binning scheme with a target state.\n")
+
+        elif nestmapname == 'BinlessMapper':
+            sim_manager = westpa.core.binning.binless_manager.BinlessSimManager(rc=self)
+
         elif drivername.lower() == 'default':
             from .sim_manager import WESimManager
 
--- a/src/westpa/core/_rc.py
+++ b/src/westpa/core/_rc.py
@@ -360,6 +360,12 @@ class WESTRC:
         elif nestmapname == 'MABBinMapper':
             we_driver = westpa.core.binning.mab_driver.MABDriver()
 
+        elif binmapname == 'BinlessMapper':
+            we_driver = westpa.core.binning.binless_driver.BinlessDriver()
+
+        elif nestmapname == 'BinlessMapper':
+            we_driver = westpa.core.binning.binless_driver.BinlessDriver()
+
         elif drivername.lower() == 'default':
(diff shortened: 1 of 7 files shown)


jpthompson17:
@atbogetti The docstrings for `map_binless` and `BinlessMapper` appear to be identical to those for `map_mab' and `MABBinMapper`. Should these be updated?

src/westpa/core/_rc.py
```diff
@@ -322,6 +322,11 @@ def new_sim_manager(self):
             from .binning.mab_manager import MABSimManager
 
             sim_manager = MABSimManager(rc=self)
+        elif binmapname == 'BinlessMapper':
```
SHZ66, line 325: Was `binmapname` defined before?

src/westpa/core/_rc.py
```diff
@@ -322,6 +322,11 @@ def new_sim_manager(self):
...
+            sim_manager = westpa.core.binning.binless_manager.BinlessManager(rc=self)
+            sys.stderr.write("-- WARNING  [config] -- Do not use BinlessMapper as an outer binning scheme with a target state.\n")
+        elif nestmapname == 'BinlessMapper':
```
SHZ66, line 328: Same as above. Depending on where these variables are from, I wonder if there is a better way of checking these conditions. It may have to use string names in the end but string names can be very unreliable.

SHZ66 (changes requested):

src/westpa/core/_rc.py
jeremyleung521: I just had a quick look at the git history. It seems like when I was resolving the conflict for `_rc.py`, I've brought back some old lines that were deleted in the latest commit. I'll push a new commit to fix that!

AnthonyBogetti:
> @atbogetti The docstrings for `map_binless()` and `BinlessMapper` appear to be identical to those for `map_mab()` and `MABBinMapper`. Should these be updated?

@jpthompson17 I am changing those now and will have them pushed soon.  Thanks for the heads-up!

atbogetti (commented):
@jeremyleung521 The BinlessMapper does not work without these lines in _rc.py that check the bin mapper name and then direct WESTPA to use the specific binless_manager.py and binless_driver.py; can we add them back in?  What was the conflict that was happening with the _rc.py?

Update: I am replicating the automatic MAB detection functions to autodetect the BinlessMapper to remove the dependency on those lines I mentioned above.  This will be part of the next round of changes I push and should solve the issue.

jeremyleung521 (approved):

PR #239, opened 2022-04-21
Add option for MAB binning in one direction only

atbogetti:
**Issue Number.** Is this pull request related to any outstanding issues? If so, list the issue number.  

**Describe the changes made.** A clear and concise description of what the problem is and what you did to fix it. E.g. [...] was happening and I've changed [...] to fix it.  

The current MAB scheme implementation will drive binning in both directions which may not always be desirable (for instance if an RMSD metric is used as a progress coordinate, driving binning towards an RMSD of zero may be counter productive). This change allows for the specification of a direction keyword that will selectively MAB bin only in the specified direction.  This can also be used in 2-D.  Please see the wiki for more information about how to use/test this.

**Goals and Outstanding Issues.** A clear and concise list of goals (to be) accomplished.  
- [x] Allow MAB binning in only 1 direction

**Major files changed.**  
- [x] mab.py

**Status.**
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the CONTRIBUTING document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

**Additional context.** Add any other context or screenshots about the pull request here.

PR #247, opened 2022-06-13
MABBinMapper Fix for restarting a partially-completed iteration

jeremyleung521:
#239 

`MABBinMapper` will assign segment twice on a restarted simulation with a partially completed iteration, once only on completed segments from the partially completed iteration, once after all of them are propagated. This will lead to trajectories with a probability of more than 1.

I've subclassed the `sim_manager` to include the `prepare_iteration()` so the BinMapper will not attempt to run assign an extra time on an incompletely simulated iteration. This didn't matter to the regular schemes because it only takes into account of the pcoord of each segment so both assignments are identical. Not the case for MAB since it depends on the pcoord of all of the segs in that iteration.

The other change is to add back code that was deleted during the MAB one-direction PR (#239). Introduced extra checks that catches segs that are mapped incorrectly due to floating point error before completely ValueError - ing out.

The `pop` > `get` change should prevent some kwargs from disappearing/changing during the simulation.

- [x] Fix issue where simulation consistently crashes with MAB after restart.

**Major files changed.**  
- [x] src/westpa/core/binning/mab.py
- [x] src/westpa/core/binning/mab_manager.py

**Status.**

### DIFF
--- a/src/westpa/core/binning/mab.py
+++ b/src/westpa/core/binning/mab.py
@@ -181,7 +181,12 @@ def map_mab(coords, mask, output, *args, **kwargs):
                     elif bin_number < 0:
                         bin_number = 0
                 elif bin_number >= nbins or bin_number < 0:
-                    raise ValueError("Walker out of boundary")
+                    if np.isclose(bins[-1], coord):
+                        bin_number = nbins - 1
+                    elif np.isclose(bins[0], coord):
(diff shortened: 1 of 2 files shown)


AnthonyBogetti (approved):
Everything is fixed now.  Thanks Jeremy!

PR #248, opened 2022-06-15
Added a MAB option to "skip" splitting in a specified pcoord dimension

AnthonyBogetti:
This PR adds an option to skip binning in a specified dimension of the progress coordinate when using MAB.  It simply checks on a per-dimension basis if the user has specified a skip and if so just assigns all walkers in that dimension to the same bin (so therefore no splitting occurs outside of that single bin). This is useful for cases where a user needs to track a value for recycling purposes but doesn't want to bin along it and incur the cost associated with an added pcoord dimension.

To use this new feature, add an option under the MABBinMapper section as follows:

`skip: [0]`

Where the list provided for skip has a value for each pcoord dimension and 0 is split as normal and 1 is skip splitting in this dimension.

**Major files changed.**  
- [ ] mab.py

**Status.**

### DIFF
--- a/src/westpa/core/binning/mab.py
+++ b/src/westpa/core/binning/mab.py
@@ -3,6 +3,8 @@ from westpa.core.binning import FuncBinMapper
 import westpa
 import logging
 
+log = logging.getLogger('westpa.rc')
+
 
 def map_mab(coords, mask, output, *args, **kwargs):
     '''Binning which adaptively places bins based on the positions of extrema segments and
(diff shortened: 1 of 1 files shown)


src/westpa/core/binning/mab.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 from westpa.core.binning import FuncBinMapper
 import westpa
+import logging
```
jeremyleung521, line 8: Add ` log = logging.getLogger('westpa.rc')`

src/westpa/core/binning/mab.py
```diff
@@ -21,6 +23,15 @@ def map_mab(coords, mask, output, *args, **kwargs):
     ndim = len(nbins_per_dim)
     if direction is None:
         direction = [0] * ndim
+    elif len(direction) != ndim:
+        direction = [0] * ndim
+        logging.warn("Direction list is not the correct dimensions, setting to defaults.")
```
jeremyleung521: Change to `log.warn` so the errors are coming from westpa.rc

src/westpa/core/binning/mab.py
```diff
@@ -21,6 +23,15 @@ def map_mab(coords, mask, output, *args, **kwargs):
...
+
+    if skip is None:
+        skip = [0] * ndim
+    elif len(skip) != ndim:
+        skip = [0] * ndim
+        logging.warn("Skip list is not the correct dimensions, setting to defaults.")
```
jeremyleung521: same as above, change to `log.warn`

jeremyleung521 (approved):
Good to merge after these changes.

PR #271, opened 2022-10-10
Ray demo

jdrusso:
Sample implementation of Ray work manager for WESTPA

Currently works
- [x] Stateless functions, like `w_assign`

Current limitations
- [ ] westpa.rc state is not passed to workers. This needs to be fixed before it can be used as a backend for stateful tasks like `w_run`


jdrusso:
Disregard, wrong remote. Github defaults PRs to the fork's upstream.

PR #274, opened 2022-10-13
Add creation of a per-segment trajectory dataset to HDF5 framework 

jdrusso:
**Issue Number.**
N/A

## **Describe the changes made.** 

Rather than storing start state structures as files on disk, allow retrieving them from HDF5-framework datasets.

This will allow completely eliminating disk IO for start-states.

This can be implemented by optionally allowing a basis_auxref for a basis/start state to be specified like `hdf5:<path/to/hdf>:iteration:seg_id` instead of a path to a file on disk. In this case, instead of looking for a file on disk to load, a structure will be loaded from the specified dataset in that h5 file.

For example, a bstates file could look like
```
01 0.1 01
02 0.1 02
03 0.1 03
04 0.1 04
05 0.1 05
06 0.1 06
07 0.1 07
08 0.1 08
09 0.1 09
10 0.05 10
11 0.05 hdf:sstates/prev_west.h5:2:4
```
where `prev_west.h5` is an h5file from a previous WE run with the HDF5 framework. (The per-iteration H5 files linked to from that dataset must be accessible.)

This functionality will be developed for the executable propagator, since the H5 framework is only directly compatible with the executable propagator.

### Overall goal:
- [ ] Allow initializing a WESTPA simulation (or, equivalently, initializing a restart) using states stored in an HDF5 file, instead of written to disk.

### Checkpoints:
- [x] Allow loading basis-states directly from an HDF5 file
- [ ] Ensure compatibility with both basis-states and start-states
- [ ] Connect to restarting plugin, so it can write structures as an H5 file instead of as individual PDBs.
     Note some extra magic will have to happen here, since each restart is synthesizing multiple H5 files. Something like...
     - [ ] Make restarting plugin write out its own h5 file, which is used in place of `prev_west.h5` in the example above, and which links `cluster_id/seg_id` to `restart/run/iter/segment`. 

## **Major files changed.**  
- [x] `propagators.executable`

## **Status.**
- [] I have added tests to cover my changes.

## **Additional context.**
N/A

### DIFF
--- a/src/westpa/core/data_manager.py
+++ b/src/westpa/core/data_manager.py
@@ -586,14 +586,15 @@ class WESTDataManager:
         if 'trajectories' not in iter_group:
             # print(f"trajectories not found in {iter_group.keys()} at iter {n_iter}")
 
+            iter_group['trajectories'] = h5py.ExternalLink(iter_ref_rel_path, '/')
             if n_iter == 0:
                 # For the ibstates in iter 0, just link to the ibstates iteration H5 file.
                 # TODO: These aren't written correctly for SynD
-                iter_group['trajectories'] = h5py.ExternalLink(iter_ref_rel_path, '/')
                 return
 
-            # TODO: Deal with this sizing in a better way!!! This is just for testing
             n_segments = self.get_iter_summary()[0]
+
+            # This feels like a slightly weird way to get these, but there's not really an option
             steps_per_seg = westpa.rc.config.get(['west', 'system', 'system_options', 'pcoord_len'])-1
             n_dims = 3
             n_frames = steps_per_seg * n_segments
@@ -603,7 +604,7 @@ class WESTDataManager:
                 dtype='<f4')
 
             iter_vsource = h5py.VirtualSource(
-                iter_ref_rel_path,
+                '.',
                 '/coordinates',
                 shape=(n_frames, n_atoms, n_dims)
             )
@@ -619,11 +620,13 @@ class WESTDataManager:
                         i * steps_per_seg:(i + 1) * steps_per_seg
                     ]
 
-            self.we_h5file.create_virtual_dataset(
-                f'{iter_group.name}/trajectories/auxdata_virtual_coord',
-                iter_layout,
-                fillvalue=-1
-            )
+            with h5py.File(iter_ref_h5_file, 'a') as outf:
+                # self.we_h5file.create_virtual_dataset(
+                outf.create_virtual_dataset(
+                    '/segment_trajectories',
+                    iter_layout,
+                    fillvalue=-1
+                )
 
     def get_basis_states(self, n_iter=None):
         '''Return a list of BasisState objects representing the basis states that are in use for iteration n_iter.'''
--- a/src/westpa/core/data_manager.py
+++ b/src/westpa/core/data_manager.py
@@ -605,27 +605,19 @@ class WESTDataManager:
             n_dims = 3
             n_frames = steps_per_seg * n_segments
 
-            iter_layout = h5py.VirtualLayout(
-                shape=(n_segments, steps_per_seg, n_atoms, n_dims),
-                dtype='<f4')
+            iter_layout = h5py.VirtualLayout(shape=(n_segments, steps_per_seg, n_atoms, n_dims), dtype='<f4')
 
-            iter_vsource = h5py.VirtualSource(
-                '.',
-                '/coordinates',
-                shape=(n_frames, n_atoms, n_dims)
-            )
+            iter_vsource = h5py.VirtualSource('.', '/coordinates', shape=(n_frames, n_atoms, n_dims))
 
-            n_remaining_segments = n_segments-1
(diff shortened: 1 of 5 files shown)


src/westpa/core/propagators/executable.py
```diff
@@ -608,17 +627,47 @@ def get_pcoord(self, state):
...
-        addtl_env, return_files, del_return_files = self.setup_dataset_return(
-            subset_keys=['pcoord', 'trajectory', 'restart', 'log']
-        )
-        addtl_env[self.ENV_STRUCT_DATA_REF] = struct_ref
+        westpa.rc.pstatus(f"getting pcoord for structure {struct_ref}")
 
-        rc, rusage = execfn(child_info, state, addtl_env)
-        if rc != 0:
-            log.error('get_pcoord executable {!r} returned {}'.format(child_info['executable'], rc))
+        state_ref_is_h5 = 'hdf:' in struct_ref
```
SHZ66: A lot of code added here seems to fit better if they were moved to the data manager. The propagators themselves typically don't interact with the files directly, instead, they pass information to the data manager through the segment object. Please take a look at `prepare_iteration` and `update_segments` and I _think_ there's a way to move this code there.

src/westpa/core/propagators/executable.py
```diff
@@ -608,17 +627,47 @@ def get_pcoord(self, state):
...
+            iteration = int(iteration)
+            seg_id = int(seg_id)
+            westpa.rc.pstatus(f"Looking for segment {seg_id} in iteration {iteration} of h5file {h5path}")
+
+            # Get the pcoord from the H5 file (assume it's formatted the same as this one!)
+            with h5py.File(h5path, 'r') as prev_h5:
 
-        self.retrieve_dataset_return(state, return_files, del_return_files, True)
+                # TODO: Get this in some non-hardcoded way
+                iter_group = f'iterations/iter_{iteration:08d}'
```
SHZ66: This is not a safe assumption. Please refer to `WESTDataManager.get_iter_group` to see how to obtain the iteration group safely. Better yet, if this code is part of the data manager instead of the propagator, you could use the data manager's functions & properties to access the data more easily.

src/westpa/core/propagators/executable.py
```diff
@@ -640,6 +689,70 @@ def gen_istate(self, basis_state, initial_state):
...
+            #   via $WEST_PARENT_DATA_REF.
+
+            # That means to prepare the first iteration, we need to make sure that's populated with valid paths.
+            # This will *probably* consist of:
+            #   1. Get all initial states
+            #   2. Remove all istate/bstates, keep only sstates
+            #   3. For each, write the structure as a file to the appropriate traj_segs segment directory, OR write it
+            #       somewhere central and set $WEST_PARENT_DATA_REF (probably better)
+            istates = self.rc.get_data_manager().get_initial_states(n_iter=1)
+            bstates = self.rc.get_data_manager().get_basis_states(n_iter=1)
```
SHZ66: It seems to me that instead of accessing the data manager here from RC, we could just move/adapt this section of code to the data manager (`prepare_iteration`).

SHZ66 (changes requested):

src/westpa/core/propagators/executable.py
jdrusso: In this section, the basis state has an HDF5 file along with segment and walker specified, *from a separate simulation*. This is not the HDF5 file generated by the current WE simulation, so the data manager has no access to this information

jdrusso:
I think rather than interacting with the H5 file directly, I can just create a *new* data manager, and then use that to interact with the H5 file

src/westpa/core/data_manager.py
```diff
@@ -572,15 +573,78 @@ def update_iter_h5file(self, n_iter, segments):
...
+
         iter_group = self.get_iter_group(n_iter)
 
         if 'trajectories' not in iter_group:
             iter_group['trajectories'] = h5py.ExternalLink(iter_ref_rel_path, '/')
 
+        # After we prepare the virtual dataset, we populate it as segments come in
+        else:
+
+            steps_per_seg = westpa.rc.config.get(['west', 'system', 'system_options', 'pcoord_len']) - 1
```
jdrusso, line 596: This is not a good way to get this, but how can we get it without explicitly reading trajectories?

src/westpa/core/data_manager.py
```diff
@@ -572,15 +573,78 @@ def update_iter_h5file(self, n_iter, segments):
...
+
+            # For the ibstates in iter 0, just link to the ibstates iteration H5 file and do nothing else.
+            if n_iter == 0:
+                return
+
+            n_segments = self.get_iter_summary()[0]
+            # TODO: Magic number, and also makes it MD-specific.. But I suppose it already is, if we're using MDtraj
+            #   trajectories to store data.
+            n_dims = 3
```
jdrusso, line 605: Similarly to the above, how can we get this without explicitly reading trajectories?

PR #384, opened 2024-03-19
Fix to make sure recursive bin labels match up with bins index

jeremyleung521:
https://groups.google.com/g/westpa-users/c/zBmbMlsXxPY

Depending on the order the recursive binning is made, `RecursiveBinMapper._recursion_target` (a dictionary of constituent bin mappers) maybe in different orders. This causes a mismatch of how [bin indices are labeled](https://github.com/westpa/westpa/blob/westpa2/src/westpa/core/binning/assign.py#L363-L382) and [how bin labels are generated](https://github.com/westpa/westpa/blob/westpa2/src/westpa/core/binning/assign.py#L352-L354). This might be a vestige of when dictionaries aren't ordered based on insertion (before py3.6).

Guarantee that the order of `RecursiveBinMapper._recursion_target` is sorted by key, which is necessary for the labels to match up.  This fix is coded this way to make sure bins are labeled in monotonically increasing order (by each dimension).

`RecursiveBinMapper._recursion_map`, which indicates if the bin is recursive, is already indexed based on position (via the key in `_recursion_target`), so should not be affected. `RecursiveBinMapper._recursion_target` is also indexed via key, so should work as expected except when it's called via `RecursiveBinMapper._recursion_target.values()` (i.e. in `RecursiveBinMapper.labels`).

Haven't tested on restarting older simulations so might be worth double checking that before merging.

- [x] Make sure bin indices line up with bin labels

**Major files changed.**  
- [x] src/westpa/core/binning/assign.py

**Status.**
- [x] Breaking change (fix or feature that would cause existing functionality to change)

Short code snippet to test. Calling `list(bin_mapper.labels)` is also useful to make sure they are in the right order.
```
from westpa.core.binning import RecursiveBinMapper, RectilinearBinMapper

def make_mapper():
    mapper_outer=RectilinearBinMapper([[0,5,float('inf')],[0, float('inf')], [0, float('inf')]])
    second_bin=RectilinearBinMapper([[5,float('inf')], [0, 1, 2, 3, float('inf')], [0, float('inf')]])
    first_bin=RectilinearBinMapper([[0,5], [0,15.0, float('inf')], [0, float('inf')]])
    
    bin_mapper = RecursiveBinMapper(mapper_outer)
    bin_mapper.add_mapper(second_bin, [10,2,5])
    bin_mapper.add_mapper(first_bin, [1, 2, 5])

    print(bin_mapper._recursion_targets)
    # should print before fix:
    # {1: <RecursiveBinMapper at 0x12061f750 with 4 bins>, 0: <RecursiveBinMapper at 0x121d122d0 with 2 bins>}
    # should print after fix:
    # {0: <RecursiveBinMapper at 0x106fc4e90 with 2 bins>, 1: <RecursiveBinMapper at 0x106f9ead0 with 4 bins>}

    return bin_mapper

def make_mapper2():
    mapper_outer=RectilinearBinMapper([[0,5,float('inf')],[0, float('inf')], [0, float('inf')]])
    second_bin=RectilinearBinMapper([[5,float('inf')], [0, 1, 2, 3, float('inf')], [0, float('inf')]])
    first_bin=RectilinearBinMapper([[0,5], [0,15.0, float('inf')], [0, float('inf')]])
    
    bin_mapper = RecursiveBinMapper(mapper_outer)
    bin_mapper.add_mapper(first_bin, [1, 2, 5])
    bin_mapper.add_mapper(second_bin, [10,2,5])

    print(bin_mapper._recursion_targets)
    # should print {0: <RecursiveBinMapper at 0x106fc4e90 with 2 bins>, 1: <RecursiveBinMapper at 0x106f9ead0 with 4 bins>}

    return bin_mapper

bm = make_mapper()
bm = make_mapper2()

```

### DIFF
--- a/src/westpa/core/binning/assign.py
+++ b/src/westpa/core/binning/assign.py
@@ -409,6 +409,9 @@ class RecursiveBinMapper(BinMapper):
 
         # we have updated our list of recursed bins, so set our own start index to trigger a recursive
         # reassignment of mappers' output values
+        # Note that we're reordering the recursion targets based on outer bin numbers (dict keys) first,
+        # so the order the mappers were added no longer matters...
+        self._recursion_targets = {k: self._recursion_targets[k] for k in sorted(self._recursion_targets)}
         self.start_index = self.start_index
 
     def assign(self, coords, mask=None, output=None):


src/westpa/core/binning/assign.py
```diff
@@ -409,6 +409,9 @@ def add_mapper(self, mapper, replaces_bin_at):
 
         # we have updated our list of recursed bins, so set our own start index to trigger a recursive
         # reassignment of mappers' output values
+        # Note that we're reordering the recursion targets based on outer bin numbers (dict keys) first,
+        # so the order the mappers were added no longer matters...
+        self._recursion_targets = {k: self._recursion_targets[k] for k in sorted(self._recursion_targets)}
         self.start_index = self.start_index
```
darianyang, line 415: Just out of curiosity, I wonder why this line is even needed?

darianyang (approved):
Ah interesting, good that we just looked at dict ordering

src/westpa/core/binning/assign.py
jeremyleung521: That's the fix (and [seemingly the fastest way to reorder a dictionary](https://stackoverflow.com/a/50569360/17047865)). Otherwise `_recursion_targets` would be in insertion order and start_index would label things differently depending on add order.

darianyang: Whoops, should have been more clear, this line: 
self.start_index = self.start_index

jeremyleung521: Probably because `start_index` is a property, and how `add_mapper` method can be run recursively, so this ensures that setter is run at every level to assign unique bin indices when each recursive mapper is created. (Otherwise `start_index` will only be set during initialization or when it's called)

darianyang: ah i see, that makes sense, I didn't realize it was a property

jeremyleung521:
Did some checks and found that no changes are done to previous simulations because those BinMappers are pickled.

Probably the best outcome since we don't want bin indices to change between versions. This PR will only affect new simulations (or more specifically RecursiveBinMapper made with the new version). Ready to be merged, I think.

PR #404, opened 2024-04-26
ibstates auxdata + ability to call "parent" segment with auxdata

jeremyleung521:
## Describe the changes made

Due to how `sim_manager.propagate()` works on `final_binning` and `sim_manager.run_we()` works on `next_iter_binning`, the `we_driver._run_we()` method have no access to auxdata. This makes coding up custom resamplers very difficult. There's also the problem where continued trajectories have auxdata and new trajectories (recycled segments, generated from an istate) do not.

Calling `Segment.parent_segment()` in any part of the workflow should now return the equivalent segment in `final_binning`. To address the discrepancy, ibstates can now have auxdata as well.

Users will have to specify `ibstates: True` in the west.cfg (`west/data/datasets`) for WESTPA to save those aux dataset for ibstates.

- [x] Quick method to get auxdata
- [x] auxdata for ibstates

## Major files changed

- [x] https://github.com/jeremyleung521/westpa/compare/725f919..9eef4ef

## Status


jpthompson17:
Another approach for preserving access to auxdata is to swap the order of resampling and recycling (#521), so resampling is done on the segments in `final_binning` instead of `next_iter_binning`.

PR #436, opened 2024-07-10
Fixing the MAB skip function in 2+ dimensions.

HScheiber:
## Issue Number
Not related to an open issue, but I have reported this in the [westpa-users mailing list](https://groups.google.com/g/westpa-users/c/qBRihv_6stA/m/HKptPU5wCAAJ).

Currently MAB skipping is broken when using 2D pcoords. Presumably it works with a 1D pcoord, but when I attempted to use it with a 2-dimensional progress coordinate and MAB skipping along particular dimensions, it failed to populate the bins properly. I believe this is due to some erroneous break statements in the loop over dimensions when skip is enabled.

## Describe the changes made

- Parsing the mab.py code was quite difficult, and it seems overly complex with extra for loops. I refactored a lot of it by renaming variables for clarity, removing duplicated code, adding comments, and converting for loops to vectorized commands where possible. I believe the code now runs significantly faster for large sets of input coords.
- I fixed the bug in skipping by refactoring the calculation of MAB bins, carefully dissecting out the erroneous break statements to work in 2D+. All code changes are contained within mab.py.
- I abstracted parts of the code into sub-functions to help with readability and maintainability.
- When `skip` is active, the code now re-assigns the number of linear bins along the skipped dimension to be `1`, irrespective of the input number of bins along that dimension. It now also properly prevents assignment of any walker to special bins along that dimension.
- `direction=86` correctly ignores leading/lagging bins and only assigns bottleneck bins (in both directions) if `bottleneck=True`. Therefore, `direction=86` with `bottleneck=False` now disables all special MAB bins along that dimension. **This is changed behavior in reporting**, as previously `bottleneck=False` reported a single bottleneck bin per dimension with `direction=86`. The behavior and bin reporting now match, ensuring this parameter works as expected.
- Notably, `maxdiff` and its flipped version are now initialized to `-np.inf` rather than 0. This way, a bottleneck bin will be found even if the largest Z is negative (notation from the MAB paper). Now, if no walker has a positive Z, a non-boundary walker with the smallest negative Z will be assigned to a bottleneck bin instead (this will typically be a near-boundary walker). **This is changed behavior**, and it makes all bottleneck bins more likely to be populated, particularly when there are few walkers within a MAB zone.

MAB now creates the expected number of bins with `direction`, `skip`, `bottleneck`, and `nbins` all playing nicely together. Here are some example configurations tested with a 2D pcoord.
**Example 1**
> direction: [0, 0]
> skip: [0, 0]
> bottleneck: true 
> nbins: [5, 1]

Produces 5x1 linear + 2x2 boundary + 2x2 bottleneck = 13 total MAB bins

**Example 2**
> direction: [1, 0]
> skip: [0, 1]
> bottleneck: true 
> nbins: [5, 1]

Produces 5 linear + 1 boundary + 1 bottleneck = 7 total MAB bins

**Example 3**
> direction: [-1, 86]
> skip: [0, 0]
> bottleneck: false
> nbins: [5, 5]

Produces 5x5 linear + 1 boundary + 0 bottleneck = 26 total MAB bins

**Example 4**
> direction: [-1, 86]
> skip: [0, 1]
> bottleneck: true
> nbins: [5, 5]

Produces 5 linear + 1 boundary + 1 bottleneck = 7 total MAB bins

- The tests should be refactored.
- MAB bin occupancy reporting may not always be accurate for bottleneck bins. This is an ongoing bug which is not addressed by these changes.

## Major files changed
- [mab.py](https://github.com/HScheiber/westpa/compare/579af04a62ab9883bfaf9468932a324acb0fbdae...d75ab2f6e66ade502a8920d76c4da117aea3b748)
- [tests/test_binning.py](https://github.com/westpa/westpa/pull/436/files#diff-c91a82fe6dbdd9f614095b620529e68361cb84883f8fd743149394946ddbec7c) (new tests added for MAB code)
- tests/ref/mab_assignments_ref.h5, which contains 2D and 3D MAB binning test expectation results.

## Status
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [X] I have read the CONTRIBUTING document.
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.

## Additional context
- [Related westpa-users discussion post 1](https://groups.google.com/g/westpa-users/c/qBRihv_6stA/m/HKptPU5wCAAJ)
- [Related westpa-users discussion post 2](https://groups.google.com/g/westpa-users/c/NFFCwBBL5Uw)


tests/test_binning.py
```diff
@@ -334,4 +340,186 @@ def test_init(self):
...
         assert mab.determine_total_bins(nbins_per_dim=[5, 1], direction=[1, 86], skip=[0, 0], bottleneck=True) == 9
+        assert mab.determine_total_bins(nbins_per_dim=[5, 5], direction=[0, 0], skip=[0, 0], bottleneck=True) == 33
+        assert mab.determine_total_bins(nbins_per_dim=[5, 5, 5], direction=[0, 0, 0], skip=[0, 0, 0], bottleneck=True) == 137
+        assert mab.determine_total_bins(nbins_per_dim=[5, 5], direction=[0, 0], skip=[1, 0], bottleneck=True) == 9
+        assert mab.determine_total_bins(nbins_per_dim=[5, 1], direction=[1, 86], skip=[0, 0], bottleneck=False) == 6
+
+        # Test bin assignments with 2D and 3D coords
+        # First load reference datasets from file
+        test_refs = []
+        with h5py.File(os.path.join(REFERENCE_PATH, 'mab_assignments_ref.h5'), 'r') as f:
```
jeremyleung521: The ref file is not included in this PR.

tests/test_binning.py
```diff
@@ -334,4 +340,186 @@ def test_init(self):
...
+        weights /= np.sum(weights)
+        allcoords = np.ones((N_total, 4))
+        allcoords[:, :2] = coords
+        allcoords[:, 2] = weights
+        allcoords = np.tile(allcoords, (2, 1))
+        allcoords[0:N_total, 3] = 0
+        mask = np.full((N_total * 2), True)
+        output = list(np.zeros((N_total * 2), dtype=int))
+
+        # TEST 1: one direction only, no bottleneck
```
jeremyleung521: Might be easier to separate all these tests into separate methods. That way it'll be easier to track which one is failing (and/or if all of them are failing). 

The reading of the reference h5file, prepping the fake dataset weights/coords from a Gaussian distribution could probably be moved into a fixture (https://github.com/westpa/westpa/blob/westpa2/tests/conftest.py).  In fact it might be cleaner to use pytest parametrize to consolidate most of the tests (https://github.com/westpa/westpa/blob/westpa2/tests/test_tools/test_w_fluxanl.py#L19-L23)

jeremyleung521 (commented):
Haven't had a change to look at the mab rewrite just yet, but I've included comments in the tests. **The main thing is that the ref file is not included in the PR.**

The reorganization of the tests, if you don't have time, I can push a PR into your branch when I have the time to do it!

HScheiber:
Okay I've correctly added in some rigorous tests. These require visualization to make sense of them, so I'll post my visualizations here.

# WESTPA Version 2022.11 MAB Test results (Before changes)

## 2D Grid test data
For the first tests, I'm using the following synthetic 2D coord data:
image

Note in the following visualizations I'm scaling the walker size by its weight and coloring by bin assignment.

### Test 1
image
**WESTPA version 2022.11 passes this test, the bins look as expected**

### Test 2
image
**WESTPA version 2022.11 fails this test.** The leading bin in the x direction and the lagging bin in the y direction end up as the same bin.

### Test 3
image
**WESTPA version 2022.11 fails this test**. MAB does not create leading/lagging bins here, instead assigning all edge coords to bottleneck bins, which is unexpected behavior. This is actually an edge case bug, normally there would not be so many coords at the same value.

### Test 4
image
**WESTPA version 2022.11 fails this test**. Similarly to the previous case version 2022.11 fails to create leading/lagging bins, and also creates too few bottleneck bins!

### Test 5
image
**WESTPA version 2022.11 fails this test**. Now we try using skipping and we get a major failure here. Not only are the leading/lagging coords missing, only a single linear bin is present in the non-skipped dimension!

### Test 6
image
**WESTPA version 2022.11 fails this test** Although the bins looks as expected, the bin ID is not 0. When skipping all bins, we would expect all coords to land in bin 0.

### Test 7 
image
**WESTPA version 2022.11 fails this test**. We are missing several of the expected special bins here, its hard to untangle what has gone wrong.

### Test 8
image
**WESTPA version 2022.11 passes this test, the bins look as expected**

## 3D Grid test data
For the next 2 tests, I'm using the following synthetic 3D coord data:
image
I assign weights from a 3D gaussian distribution.

In these visualizations I'm coloring by bin assignment but not scaling walker size by weight.

### Test 9
image
**WESTPA version 2022.11 fails this test.** Some sides are not assigned unique leading/lagging bins correctly.

### Test 10
image
**WESTPA version 2022.11 fails this test.** Multiple things go wrong here, far fewer special bins are assigned than expected.

## 2D Gaussian test data
For the last set of tests, I'm using the following pseudorandom synthetic 2D coord data:
image

Note in these visualizations I'm scaling the walker size by its weight and coloring by bin assignment.

### Test 11
image
**WESTPA version 2022.11 fails this test.** Some of the directional bins are incorrectly assigned!

### Test 12
image
**WESTPA version 2022.11 fails this test.** This again demonstrates how broken the skip function is in 2D.

In the following post I will show the same results after making the new MAB changes in this PR.

HScheiber:
Here are the same MAB tests performed after the changes in this PR

# WESTPA MAB Test results (After changes)

## 2D Grid test data
For the first tests, I'm using the following synthetic 2D coord data:
image

Note in these visualizations I'm scaling the walker size by its weight and coloring by bin assignment.

### Test 1
image
**New MAB code passes this test, the bins look as expected**

### Test 2
image
**New MAB code passes this test, the bins look as expected**

### Test 3
image
**New MAB code passes this test, the bins look as expected**. Note how each edge now has only a single bottleneck walker, and how the leading/lagging bins are also present.

### Test 4
image
**New MAB code passes this test, the bins look as expected**.

### Test 5
image
**New MAB code passes this test, the bins look as expected**. Skipping now works!

### Test 6
image
**New MAB code passes this test, the bins look as expected**.  Everything is assigned to bin 0 correctly.

### Test 7 
image
**New MAB code passes this test, the bins look as expected**. 

### Test 8
image
**New MAB code passes this test, the bins look as expected**. 

## 3D Grid test data
For the next 2 tests, I'm using the following synthetic 3D coord data:
image
I assign weights from a 3D gaussian distribution.

In these visualizations I'm coloring by bin assignment but not scaling walker size by weight.

### Test 9
image
**New MAB code passes this test, the bins look as expected**. It's hard to tell but the bins are correctly assigned in 3D as well.

### Test 10
image
**New MAB code passes this test, the bins look as expected**. Again, you might have to trust me but the bins are assigned correctly, now with bottleneck bins too.

## 2D Gaussian test data
For the last set of tests, I'm using the following pseudorandom synthetic 2D coord data:
image

Note in these visualizations I'm scaling the walker size by its weight and coloring by bin assignment.

### Test 11
image
**New MAB code passes this test, the bins look as expected**.

### Test 12
image
**New MAB code passes this test, the bins look as expected**. Skipping now works with realistic data as well.

tests/test_binning.py
HScheiber: Sorry it got blocked by gitignore. I added it in a subsequent commit.

tests/test_binning.py
HScheiber: Will look into rearranging the tests when I have more time, but they are at least all passing at the moment.

HScheiber:
On closer inspection, Test 7 actually failed after the changes I had previously made! I have gone in and corrected that bug and refactored the offending code, simplifying it greatly. 

### Test 7 before the latest commit:
image

### Test 7 after the latest commit:
image

I also refactored the tests, added an additional test, and added the code I used to generate the reference data into the tests under a `if __name__ == 'main'` block.

tests/test_binning.py
HScheiber: I've refactored the tests as per your suggestions.

jeremyleung521 (approved):
LGTM. Sorry for the wait. Someone else is also currently testing this and we'll aim for final approval/merging this PR next week.

gma57 (approved):
Looks good to me! Tested on 2D ODLD potential.

PR #452, opened 2024-10-15
Allow WESS and WEED to detect when a simulation has been truncated

gma57:
## Describe the changes made
Previously, when w_truncate was used to truncate a simulation, WESS and WEED would not execute reweighting until the simulation progressed beyond the original last reweighting iteration. This is because they used an attribute field in the west.h5 to store the last reweighting iteration, which would overwrite each time reweighting occurs. This value would stay static after truncating the simulation, leading to undesired behavior of WESS and WEED not reweighting at all after truncating, since the last reweighting occurred in a future iteration.

To fix this, I instead stored the reweighting history data in a dataset. This is more appropriate since attributes should be used for metadata, and it allows a running history of reweighting events to be recorded. I also added a check to evaluate if the last reweighting event is in a future iteration. If this is the case, the reweighting history dataset will be truncated to only include values less than n_iter.

- [x] Made WESS and WEED compatible with w_truncate

## Major files changed
- [x] src/westpa/westext/weed/weed_driver.py
- [x] src/westpa/westext/wess/wess_driver.py

## Status

### DIFF
--- a/src/westpa/westext/weed/weed_driver.py
+++ b/src/westpa/westext/weed/weed_driver.py
@@ -101,7 +101,7 @@ class WEEDDriver:
             last_reweighting = int(weed_global_group.attrs.get('last_reweighting', 0))
             if last_reweighting > n_iter:
                 last_reweighting = n_iter - 1
-                reweighting_history = np.array(reweighting_history_dataset)
+                reweighting_history = reweighting_history_dataset[:]
                 reweighting_history = reweighting_history[reweighting_history < n_iter]
                 reweighting_history_dataset.resize((reweighting_history.size), axis=0)
 
--- a/src/westpa/westext/weed/weed_driver.py
+++ b/src/westpa/westext/weed/weed_driver.py
@@ -94,14 +94,16 @@ class WEEDDriver:
             return
 
         with self.data_manager.lock:
-            weed_global_group = self.data_manager.we_h5file.require_dataset('weed', (1,), maxshape=(None,), dtype=int)
-            last_reweighting = int(weed_global_group[-1])
-
-        if last_reweighting > n_iter:
-            reweighting_history = np.array(weed_global_group)
-            reweighting_history = reweighting_history[reweighting_history < n_iter]
-            weed_global_group.resize((reweighting_history.size), axis=0)
-            weed_global_group[:] = reweighting_history
+            weed_global_group = self.data_manager.we_h5file.require_group('weed')
+            reweighting_history_dataset = weed_global_group.require_dataset(
+                'reweighting_history', (1,), maxshape=(None,), dtype=int
+            )
+            last_reweighting = int(weed_global_group.attrs.get('last_reweighting', 0))
+            if last_reweighting > n_iter:
+                last_reweighting = n_iter - 1
+                reweighting_history = np.array(reweighting_history_dataset)
+                reweighting_history = reweighting_history[reweighting_history < n_iter]
+                reweighting_history_dataset.resize((reweighting_history.size), axis=0)
 
         if n_iter - last_reweighting < self.reweight_period:
             # Not time to reweight yet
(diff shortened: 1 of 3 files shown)


src/westpa/westext/weed/weed_driver.py
```diff
@@ -94,8 +94,14 @@ def prepare_new_iteration(self):
...
-            weed_global_group = self.data_manager.we_h5file.require_group('weed')
-            last_reweighting = int(weed_global_group.attrs.get('last_reweighting', 0))
+            weed_global_group = self.data_manager.we_h5file.require_dataset('weed', (1,), maxshape=(None,), dtype=int)
+            last_reweighting = int(weed_global_group[-1])
+
+        if last_reweighting > n_iter:
+            reweighting_history = np.array(weed_global_group)
+            reweighting_history = reweighting_history[reweighting_history < n_iter]
+            weed_global_group.resize((reweighting_history.size), axis=0)
+            weed_global_group[:] = reweighting_history
```
jeremyleung521: You're creating a completely new array, then trimming that, then setting that as the new "wess" dataset (should be a group, see above, so double check), which will bloat up the size of the file. Wouldn't resizing be sufficient to remove the data from reference? (you'd need to repack the file to reclaim the space, but that's up to the user)

src/westpa/westext/weed/weed_driver.py
```diff
@@ -94,8 +94,14 @@ def prepare_new_iteration(self):
             return
 
         with self.data_manager.lock:
-            weed_global_group = self.data_manager.we_h5file.require_group('weed')
-            last_reweighting = int(weed_global_group.attrs.get('last_reweighting', 0))
+            weed_global_group = self.data_manager.we_h5file.require_dataset('weed', (1,), maxshape=(None,), dtype=int)
```
jeremyleung521: You turned a group into a dataset, which are different types of data in hdf5. Doesn't seem like this is ok for backwards compatibility?

src/westpa/westext/weed/weed_driver.py
```diff
@@ -172,7 +178,8 @@ def prepare_new_iteration(self):
             for bin, newprob in zip(bins, binprobs):
                 bin.reweight(newprob)
 
-            weed_global_group.attrs['last_reweighting'] = n_iter
+            weed_global_group.resize((weed_global_group.shape[0] + 1), axis=0)
```
jeremyleung521: The `last_reweighting` attribute is gone. Is that ok? Wouldn't it be better to have a separate dataset for the history and keep the last_reweighting around?

src/westpa/westext/weed/weed_driver.py
gma57: Removed the line reassigning the values in the update.

src/westpa/westext/weed/weed_driver.py
gma57: Changed this so that the dataset is within the 'wess'/'weed' group.

src/westpa/westext/weed/weed_driver.py
gma57: Reinstated the last_reweighting attribute.

src/westpa/westext/weed/weed_driver.py
```diff
@@ -95,7 +95,15 @@ def prepare_new_iteration(self):
 
         with self.data_manager.lock:
             weed_global_group = self.data_manager.we_h5file.require_group('weed')
+            reweighting_history_dataset = weed_global_group.require_dataset(
+                'reweighting_history', (1,), maxshape=(None,), dtype=int
+            )
             last_reweighting = int(weed_global_group.attrs.get('last_reweighting', 0))
+            if last_reweighting > n_iter:
+                last_reweighting = n_iter - 1
+                reweighting_history = np.array(reweighting_history_dataset)
```
jeremyleung521: I would just change this line to `reweighting_history = reweighting_history_dataset[:]` Less of a performance hit.

src/westpa/westext/wess/wess_driver.py
```diff
@@ -112,7 +112,15 @@ def prepare_new_iteration(self):
 
         with self.data_manager.lock:
             wess_global_group = self.data_manager.we_h5file.require_group('wess')
+            reweighting_history_dataset = wess_global_group.require_dataset(
+                'reweighting_history', (1,), maxshape=(None,), dtype=int
+            )
             last_reweighting = int(wess_global_group.attrs.get('last_reweighting', 0))
+            if last_reweighting > n_iter:
+                last_reweighting = n_iter - 1
+                reweighting_history = np.array(reweighting_history_dataset)
```
jeremyleung521: I would just change this line to `reweighting_history = reweighting_history_dataset[:]` Less of a performance hit.

jeremyleung521 (approved):
Otherwise LGTM.

Issue #456, opened 2024-11-06
w_init failing to save trajectory, restart, and log data to the per-iteration HDF5 file at iteration 0.


HScheiber:
## Prerequisites
* [x] Can you reproduce the problem in safe mode?
* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

## Describe the bug

When using per-iteration HDF5 files, I found that w_init will not properly generate the per-iteration HDF5 file for iteration 0 if using the processes work manager, but it will correctly produce the HDF5 file if using serial or threads work manager with a single worker.

## To Reproduce
Enable per-iteration HDF5 files, use the processes work manager, and use the following get_pcoord.sh for your simulation.

- pcoords.txt is just a pre-computed text file with the pcoords of the basis states
- seg.h5 is a trajectory file saved with mdtraj for your system of interest.
- seg.trr is a trajectory file containing just one frame (multiple frames will work as well), corresponding to the microstate of your basis state. This was generated using GROMACS, but I'm pretty sure this bug is data type agnostic.

```shell
#!/bin/bash
if [ -n "$SEG_DEBUG" ] ; then
  set -x
  env | sort
fi

# Return the progress coordinate value for this segment
cat $WEST_BSTATE_DATA_REF/pcoords.txt > $WEST_PCOORD_RETURN

# Now we save the solute trajectory into the HDF5 framework for storage
cp $WEST_BSTATE_DATA_REF/seg.h5 $WEST_TRAJECTORY_RETURN/seg.h5

# This is the only restart file actually needed by gromacs for the next iteration
# Note the seg.trr trajectory file is a full precision file with solute and solvent
# The seg.gro file is not needed as a common file is used for all segments
# An input seg.edr file is not read by gmx grompp unless using the Nose-Hoover thermostat 
# and/or Parinello-Rahman barostat
cp $WEST_BSTATE_DATA_REF/seg.trr $WEST_RESTART_RETURN/parent.trr
```

## Expected behavior
After running 
```shell
w_init $BSTATE_ARGS $TSTATE_ARGS --work-manager=processes --n-workers=48 "$@" >> west.log 2>&1
```
I would expect a per-iteration HDF5 file containing the trajectory and restart data for my basis states to be present in `$WEST_SIM_ROOT/trag_segs/iter_000000.h5`.

## Actual behavior
`$WEST_SIM_ROOT/trag_segs/iter_000000.h5` is created by w_init, but it is empty

[...]

PR #457, opened 2024-11-06
Fix w_init when using hdf5 framework with processes work manager

jeremyleung521:
## Issue Number

#456 

## Describe the changes made

Copy the entire BasisState object from work manager during pcoord calculations instead of just the pcoord.

- [x] Have w_init work correctly with hdf5 framework + processes work manager

## Major files changed

- [x] src/westpa/core/sim_manager.py

## Status

### DIFF
--- a/src/westpa/core/sim_manager.py
+++ b/src/westpa/core/sim_manager.py
@@ -235,7 +235,7 @@ class WESimManager:
         futures = [self.work_manager.submit(wm_ops.get_pcoord, args=(basis_state,)) for basis_state in basis_states]
         fmap = {future: i for (i, future) in enumerate(futures)}
         for future in self.work_manager.as_completed(futures):
-            basis_states[fmap[future]].pcoord = future.get_result().pcoord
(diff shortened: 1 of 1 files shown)


HScheiber (approved):
While debugging this, I checked that the `future.get_result()` object is otherwise the same as the `basis_states[fmap[future]]` and found that the only difference between these objects is the pcoord and the data. Everything else should be identical, so it should be safe make this change.

I suggest modifying the docstring of this function to account for the change.

jeremyleung521:
Just leaving some notes/thoughts here for documentation purposes:
Apparently in another branch I copied the data and pcoord separately, but I think copying the whole `BasisState` object over would make it cleaner for Python garbage collection, considering that the work manager probably already transferred the whole `BasisState` object over to the main process during `future.get_results()`. 

https://github.com/jeremyleung521/westpa/pull/29/files#diff-72c2c83ebfdf534badc6cf38196b1815d12c290ac19965c88616ccb285d6b79aR238-R240

Bailyford (approved):
Looks good.

Issue #467, opened 2025-02-01
Define the sink without reference to the bin mapper


jpthompson17:
## Is your feature request related to a problem? Please describe

Currently, the definition of the sink depends on the bin mapper (i.e., the sink is the union of bins containing target states). This dependence complicates the use of adaptive binning methods in nonequilibrium steady state simulations, since it's usually desirable for the sink to remain constant during the simulation. (A current solution is to [embed the adaptive bin mapper in a recursive bin mapper with a fixed base mapper](https://github.com/westpa/westpa2_tutorials/blob/2640a38f0f104060151717a1bcea443ffff3ff66/tutorial7.6-membrane/west.cfg#L36), but this feels a bit clunky to me.) Moreover, the dependence on the bin mapper unnecessarily couples *recycling* to *resampling*, and gets in the way of implementing truly binless resampling methods.

## Describe the solution you'd like

Instead of defining the sink indirectly via a bin mapper and one or more target states:

`west.cfg`:
```yaml
      bins:
        type: RectilinearBinMapper
        boundaries:         
          - [0.0, 0.6, 0.8, 1.0, 1.2, 1.4, 1.6, 1.8, 2.0, 2.2, 2.4, 2.6, 2.8, 3.0, 3.2, 3.4, 3.6, 3.8, 4.0, .inf]
```
`tstate.file`:
```
unfolded 4.02
```

I'd like to be able to define the sink directly by specifying a region in progress coordinate space, e.g.,

`west.cfg`:
```yaml
      sink: 'x: x[0] > 4.0'
```

jpthompson17:
@jeremyleung521 While working to make [these changes](https://github.com/westpa/westpa/compare/develop...jpthompson17:westpa:feature/467-sink-definition) compatible with the CLI tools, I noticed that `w_bins` prints the number of target states, e.g.,
```
22 bins total, 1 targets, 8 (38.10%) occupied
```
Since information about *states* is arguably outside the scope of `w_bins`, I was wondering if we could remove the number of target states from the output, e.g.,
```
22 bins total, 8 (38.10%) occupied
```
This would make it easier to handle the case where the user defines the sink directly (as above), without specifying any target states.

jeremyleung521:
Yea, that sounds fine to me. Since the target state will no longer be a "bin" anymore.

PR #477, opened 2025-03-25
More flexible sink definition

jpthompson17:
## Issue Number
#467 

## Describe the changes made
Added a new `sink` configuration option. The value of `sink` is a string of the form `'<variables> : <predicate>'` (as in set builder notation), for example:
```yaml
west:
  system:
    system_options:
      sink: 'distance : distance < 2.6'
```
Here `<variables>` is a comma-separated list of variable names that label consecutive axes of progress coordinate space (no commas are necessary for the 1D case), and `<predicate>` is a boolean expression indicating whether a given progress coordinate value falls in the sink.

Use of the `sink` parameter overrides any sink(s) defined by target states.

- [x] Implement new `Sink` class.
- [x] Integrate new sink definition into `core`, `analysis`, and `cli.tools`.
- [x] Document new `sink` dataset in `west.h5` file schema docs. 
- [ ] Write comprehensive tests for changes (partially done in `test_sink.py`).
- [ ] Document new `sink` option. (Where? Is there a general description of the `west.cfg` schema in the user docs?)

## Major files changed
https://github.com/westpa/westpa/compare/develop...jpthompson17:feature/467-sink-definition

## Status


jeremyleung521 (commented):
Tested this and the basic functionality works. Though I do think it needs some small polishing regarding formulating syntax of the selection so it's more in line with the rest of the `west.cfg`. We were thinking like dataset names for the each dimension of the pcoord, which suggests it might be worth pushing this to WESTPA 3.0.

jpthompson17:
Regarding names for progress coordinate dimensions, we could add a `pcoord_dims` option that defaults to `['pcoord0', 'pcoord1', ...]`, then require only the `<predicate>` part in `sink`. The example above could then be written as
```yaml
west:
  system:
    system_options:
      sink: pcoord0 < 2.6
```
or, if the user wanted to name the progress coordinate explicitly,
```yaml
west:
  system:
    system_options:
      pcoord_dims:
      - distance
      sink: distance < 2.6
```
We could also infer `pcoord_ndim` from `pcoord_dims`, if only the latter were provided.

jeremyleung521:
> Regarding names for progress coordinate dimensions, we could add a `pcoord_dims` option that defaults to `['pcoord0', 'pcoord1', ...]`, then require only the `<predicate>` part in `sink`. The example above could then be written as
> 
> ```yaml
> west:
>   system:
>     system_options:
>       sink: pcoord0 < 2.6
> ```
> 
> or, if the user wanted to name the progress coordinate explicitly,
> 
> ```yaml
> west:
>   system:
>     system_options:
>       pcoord_dims:
>       - distance
>       sink: distance < 2.6
> ```
> 
> We could also infer `pcoord_ndim` from `pcoord_dims`, if only the latter were provided.

This sounds good. I'm also worried how some tests worked and some failed, so maybe that needs some fixing.

jpthompson17:
In the new Python API (#518), the current plan is to have a `Sink` type parametrized by an arbitrary indicator function on "segment space", e.g.,
```py
import westpa
sink = westpa.Sink(indicator=lambda segment: segment.pcoord[-1, 0] < 2.6)
```
This has the advantage of more flexibility (including access to auxiliary data) and less maintenance (no parser).

Since it seems likely we'll go in a different direction, I'm closing this PR.

PR #478, opened 2025-03-31
w_direct/w_kinetics outputfile inheritance + w_assign output folder creation

jeremyleung521:
## Describe the changes made

`w_direct` is a wrapper around various older kinetics tools (e.g. `w_kinetics`), but it is not coded in a way such that the `self.output_filename` is correctly passed to those kinetic tools as `self.kinetics_filename`. This will lead to a `FileNotFoundError` when not using the default output_filename (i.e. not `direct.h5`) via variations of the following command: `w_direct [any subtool]  -o [NOT-DEFAULT-FILE-NAME].h5`. 

In other words, I'm fixing what this inline comment is pointing out:
```
# One minor issue; as this stands now, since it's inheriting from all the other classes, it needs
# a kinetics file to instantiate the other attributes.  We'll need to modify how the loading works, there.
```
https://github.com/jeremyleung521/westpa/blob/develop/src/westpa/cli/tools/w_direct.py#L501-L502

Also bundled in a fix to cases where `w_assign` will error out with `FileNotFoundError` when the `ANALYSIS` folder is made but the specific scheme folder is not.

- [x] Have `w_direct` respect user-provided output file name

## Major files changed

- [x] src/westpa/cli/tools/w_direct.py
- [x] src/westpa/tools/kinetics_tool.py
- [x] src/westpa/cli/tools/w_assign.py

## Status

## Additional context

Try running:
`w_direct all -o NON-DEFAULT-FILE.h5`.

### DIFF
--- a/src/westpa/cli/tools/w_assign.py
+++ b/src/westpa/cli/tools/w_assign.py
@@ -340,14 +340,20 @@ Command-line options
         from westpa.core._rc import bins_from_yaml_dict
 
         self.binning.mapper = bins_from_yaml_dict(config['analysis_schemes'][scheme]['bins'][0])
+
+        # Output file/folder wrangling
         path = os.path.join(os.getcwd(), config['directory'], scheme)
+        self.output_filename = os.path.join(path, 'assign.h5')
+
         try:
             os.mkdir(config['directory'])
-            os.mkdir(path)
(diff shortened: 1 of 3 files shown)


Bailyford (approved):
Looks good.

Issue #294, opened 2023-01-23
Feature Request: Speeding up the HDF5 Framework


jeremyleung521:
**Is your feature request related to a problem? Please describe.** A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]  
The HDF5 Framework currently collects all the restart/coordinate files from the RETURNs and process all the segments serially in the server node. This results in cases where it's **_very, very_** slow because a) You're running serially through hundreds of segments potentially on a single CPU/core b) MDTraj is creating a `Trajectory` object for each segment, which contains plenty of (potentially useless) bells and whistles.

**Describe the solution you'd like.** A clear and concise description of what you want to happen.  
A) Most of the `MDTraj.Trajectory` object (like the topology) does not change between segments. Have a "dummy" template trajectory where the xyz and box (and other changes) are replaced. This would save tons of time creating a new Trajectory object each time from scratch. Would speed things up because, again, MDTraj is **_slow_** and you do this hundreds of times.

B) Instead of leaving the files to be processed by the server node, have each worker node work on the coordinates before terminating, sending the coordinates as part of segment['data']. Free Parallelization within the current server/worker framework!

C) Since it's an array of coords (and maybe the box dimensions) being passed back, users can choose to customize their own coordinate extraction process in runseg.sh. I've had success ripping the necessary coordinates/box from the netcdf files directly and it's much quicker.

**Describe alternatives you've considered.** A clear and concise description of any alternative solutions or features you've considered.  
I don't expect much to change for the restart (since it's just tarred/untarred as needed). I assume much of this can be worked in tandem with the eventual MDTraj streamlining/dependency-sunsetting process.

**Additional context.** Add any other context or screenshots about the feature request here.  
Would be great if @SHZ66 could comment on the feasibility. Just documenting a potential point-of-improvement for the future.

SHZ66:
@jeremyleung521 Thanks a lot for the post and giving it a lot of thought! I think they all make sense to me overall. Here's my thoughts on the feasibility for each point:
A) This should be fairly easy to be implemented. Instead of the full trajectory object, each segment can just store the xyz coordinates in their data, and then the trajectory object is recovered when the segments were collected. How and where to create the template might require a bit of thinking but it should be feasible.

[...]

PR #484, opened 2025-04-16
Speeding up HDF5 Framework

jeremyleung521:
## Issue Number

#294 

## Describe the changes made

In addition to [actually speeding up MDTraj loading](https://github.com/mdtraj/mdtraj/pull/2012), I've modified the HDF5 Framework such that coordinate extraction could be offloaded to the workers or be done by MDAnalysis. Topology loading in every iteration could also be bypassed by copying a template file with only the topology to speed up trajectory saving into the per-iter H5 File. (Except for the first time needed to create that template, either in `w_init` or `w_run`)

Previous behavior (mdtraj) is retained as a fallback. Users may use `mdtraj` or `mdanalysis` or `netcdf4` (for AMBER `.nc` files) for the coordinate extraction step, which could also be done manually in `runseg.sh`. 

### Sample directory
https://github.com/jeremyleung521/westpa-test-systems/tree/hdf5-amber-test/test-hdf5-amber

#### LET WESTPA EXTRACT COORDS FROM TRAJECTORIES
Users may also use `MDAnalysis`, `netCDF4`, scipy.io.netcdf_file for all the trajectory loading by setting the following in `west.cfg`. Users should expect the same behavior and would pass the same files to `$WEST_TRAJECTORY_RETURN` as before.
```
west:
  executable:
    datasets:
      - name:   trajectory
        loader:  mdanalysis_trajectory_loader
```

loader OPTIONS:
```
mdanalysis_trajectory_loader: use MDAnalysis
mdtraj_trajectory_loader: use MDTraj  # default, same as previous behavior
netcdf_trajectory_loader: use NetCDF4 to load trajectories. # for Amber .nc trajectories only
npy_data_reader: extract coordinates manually.  # Pass a .npy file to $WEST_TRAJECTORY_RETURN instead.
```

#### IF YOU WANT TO EXTRACT COORDINATES FROM TRAJECTORIES YOURSELF:
- `get_coords.py`, script for extracting coordinates
```
from netCDF4 import Dataset
import numpy
rootgrp = Dataset('parent.ncrst', 'r', format="NETCDF3")
rootgrp2 = Dataset('seg.nc', 'r', format="NETCDF3")

coords = numpy.concatenate(([rootgrp.variables['coordinates'][:]], rootgrp2.variables['coordinates'][:]))

numpy.save('coords.npy', coords)
```

- In `runseg.sh`, extract and pass coordinates (in a npy pickle file) to `$WEST_TRAJECTORY_RETURN`
```
ln -sv $WEST_SIM_ROOT/common_files/get_coords.py
# Scrape coords and output into npy file.
python get_coords.py
cat coords.npy > $WEST_TRAJECTORY_RETURN
```

- Set trajectory to use `numpy_data_loader`
```
west:
  executable:
    dataset:
      - name:    trajectory
        loader:   numpy_data_loader
```

- [x] Generalizeable loaders for trajectory saving (mdtraj, mdanalysis and amber/netcdf)
- [x] Template file for trajectory saving (to skip unnecessary mdtraj topology read/writes)
- [x] Convert MDAnalysis topology to MDTraj topology
- [x] Lessened Memory Usage
- [x] Lessened Run time
- [ ] Create WESTTrajectory class without subclassing `mdtraj.Trajectory` or `MDTraj.Topology`

## Major files changed

- [x] src/westpa/core/data_manager.py
- [x] src/westpa/core/h5io.py
- [x] src/westpa/core/trajectory.py

## Status


src/westpa/core/propagators/loaders.py
```diff
@@ -0,0 +1,185 @@
...
+    Please see ``load_netcdf`` for more details. ``single_point`` is not used by this loader.'''
+    try:
+        data = load_netcdf(coord_folder)
+        segment.data['iterh5/trajectory'] = data
+    except Exception as e:
+        log.warning('Falling back to default loader for {}: {}'.format(fieldname, str(e)))
+        mdtraj_trajectory_loader(fieldname, coord_folder, segment, single_point)
+
+
+def mda_trajectory_loader(fieldname, coord_folder, segment, single_point):
```
jpthompson17: For the sake of readability, could we spell out `mdanalysis` instead of using `mda`?

src/westpa/core/trajectory.py
```diff
@@ -321,10 +453,132 @@ def load_trajectory(folder):
...
+    return traj
+
+
+def load_netcdf(folder):
+    '''Load netcdf file from ``folder`` using ``scipy.io`` and return a ``mdtraj.Trajectory``
+    object. The folder should contain a Amber trajectory file with extensions `.nc` or `.ncdf`.
+
+    Note coordinates and box lengths are all divided by 10 to change from Angstroms to nanometers.
+    '''
+    from scipy.io import netcdf_file
```
jpthompson17, line 533: Could we put [imports at the top of the file](https://peps.python.org/pep-0008/#imports) in new code?

devtools/conda-envs/test_env.yaml
```diff
@@ -15,6 +15,8 @@ dependencies:
   - mpi4py
   - tqdm
   - mdtraj
+  - netCDF4
```
jpthompson17: `netCDF4` -> `netcdf4`
`MDAnalysis` -> `mdanalysis`

pyproject.toml
```diff
@@ -45,7 +45,6 @@ dependencies = [
   "blessings",
   "ipykernel",
   "tqdm",
-  "pandas",
```
jpthompson17, line 48: Why remove pandas as a dependency? If we must, could we wait until the next major release, since the removal breaks `westpa.analysis`?

src/westpa/analysis/core.py
```diff
@@ -25,6 +30,11 @@ class Run:
 
     def __init__(self, h5filename='west.h5'):
         self.h5filename = h5filename
+        if pd is None:
+            warnings.warn(
+                "A lot of `westpa.analysis` rely on optional dependency `pandas`. Note that "
+                "some functions (e.g., summaries) might not work since you do not have `pandas` installed."
+            )
```
jpthompson17: The summary methods _definitely_ won't work if pandas is not installed. If we're going to drop pandas (a break I think should only happen with a major release), then we should rewrite these methods.

src/westpa/core/data_manager.py
```diff
@@ -571,13 +586,35 @@ def update_iter_h5file(self, n_iter, segments):
             return
 
         west_h5_file = makepath(self.we_h5filename)
-        iter_ref_h5_file = makepath(self.iter_ref_h5_template, {'n_iter': n_iter})
+        iter_ref_h5_file = makepath(self.iter_ref_h5_path_template, {'n_iter': n_iter})
         iter_ref_rel_path = relpath(iter_ref_h5_file, dirname(west_h5_file))
+        if self.iter_ref_h5_template:
+            # Make path to per-iter H5 File
+            iter_ref_h5_file_template = makepath(self.iter_ref_h5_template, {'n_iter': n_iter})
```
jpthompson17: Does `self.iter_ref_h5_template` still have a `{n_iter}` field to replace?

Given the [example](https://github.com/westpa/westpa2_tutorials/blob/0ae27946e95f53b654c897114cc849f8b0df8bf3/tutorial7.5-hdf5/west.cfg#L40C22-L40C67) in the HDF5 framework tutorial, i.e.,
```yaml
west:
  data:
    data_refs:
      iteration:  $WEST_SIM_ROOT/traj_segs/iter_{n_iter:06d}.h5
```
it seems like `self.iter_ref_h5_template` would be `'$WEST_SIM_ROOT/traj_segs/iter_template.h5'` (after the `re.sub()` replacement). 

On a more general note, I find it a bit hard to distinguish between `iter_ref_h5_path_template` and `iter_ref_h5_template` when reading the code. Could we chose more distinct variable names?

src/westpa/core/propagators/executable.py
```diff
@@ -191,6 +58,21 @@ class ExecutablePropagator(WESTPropagator):
     ENV_RAND128 = 'WEST_RAND128'
     ENV_RANDFLOAT = 'WEST_RANDFLOAT'
 
+    @staticmethod
```
jpthompson17: Since we're using the `makepath` function in `data_manager.py`, would it make sense to define it at the module level, then bind it to `ExecutablePropagator` as a static method?
```py
def makepath(...):  # perhaps this could be in data_manager.py?
    ...


class ExecutablePropagator:
    makepath = staticmethod(makepath)
```
It looks a bit weird to me to be relying on a method of `ExecutablePropagator` in the data manager code.

src/westpa/core/trajectory.py
```diff
@@ -1,16 +1,113 @@
...
+    for atom in residue.findall('Atom'):
+        name = atom.attrib['name']
+        for id in atom.attrib:
+            map[atom.attrib[id]] = name
+
+
+def loadNameReplacementTables():
+    '''Load the list of atom and residue name replacements. Taken from OpenMM 8.2.0.'''
+
+    # importing things here because they're only used in this function
```
jpthompson17, line 21: PEP 8 recommends that all imports be at the top of the file.

src/westpa/data/pdbNames.xml
```diff
@@ -0,0 +1,283 @@
+<!--Taken from OpenMM 8.2.0, which is also under MIT License-->
```
jpthompson17, line 1: Carrying around all this MD-specific data seems like a maintenance burden and outside the scope of WESTPA. Is it absolutely necessary?

src/westpa/core/trajectory.py
jeremyleung521: This was mostly a leftover from in an older version, I used netCDF4 (and that was never going to be a hard dependency.) Will move this up.

src/westpa/data/pdbNames.xml
jeremyleung521: It is necessary if we are to accept both mdtraj and mdanalysis as options while not having either as "hard dependency" (and transfer between them to retain our current per-iter hdf5 file structure). 

This file itself hasn't changed much--atom names don't change very often. [Mdtraj's](https://github.com/mdtraj/mdtraj/blob/main/mdtraj/formats/pdb/data/pdbNames.xml) had been working without problem for 9 years. Most edits on openmm ([which doesn't happen often](https://github.com/openmm/openmm/commits/master/wrappers/python/openmm/app/data/pdbNames.xml)) add alternate namings from other programs.

src/westpa/core/trajectory.py
jeremyleung521: As much as I'd like to follow PEP8, I tried moving the imports up and down a few times and still feel strongly that placing the imports per-function is the best way (i.e. ignoring PEP8 on good faith here).  It's cluttering the file with 10-20 extra lines up top sacrificing readibility, on imports that might only every be imported/run once if you happen to use `mdanalysis` to create the h5 template file. I don't think it's worth the startup cost to load any of these unless deemed necessary (i.e. when the convert function, defined right after this function, is called). I added `@functools.cache` to relevant functions make sure we memoize the outputs so we never have to run these after the first time.

The `mdtraj`, `mdanalysis`, `scipy.io.netcdf_file` ones are in the specific loader function because the former two will eventually be optional dependencies. The latter is just trying to follow the same format as the first two (also added benefit towards making each loader independent). All loader-related imports are unnecessary except for the chosen loader you're going to use, afterall.


Doing benchmark:
Locally on an PCIE Gen 5 nvme ssd/high spec machine, skipping imports saves us ~0.26s from load time, and every subsequent reload only takes < 20 us. This will be higher on slow/high latency/busy clusters. I tried it on a login node on h2p and it took 2+ seconds (30s on the first try, even).

Locally, wrapping them in try/excepts will still take 0.17s if raising ModuleNotFoundError (and ~0.70us per reload), so the best case is not loading at all and putting them in the functions that are often never used. (i.e we'll save the entire  time taken).  Python did recently pass a PEP on lazy loading (https://peps.python.org/pep-0810/) so it's definitely a problem.

tests/test_executable.py
```diff
@@ -101,6 +101,8 @@ def test_seglog_loader_writer(self, nacl_restart_files):
         # Write the current file into self.write_dir
         seglog_writer(self.write_dir, test_segment)
 
+        print(self.write_dir)
```
jpthompson17: Should this `print()` be removed?

src/westpa/core/data_manager.py
```diff
@@ -249,8 +267,17 @@ def process_config(self):
             ['west', 'data', 'aux_compression_threshold'], self.default_aux_compression_threshold
         )
         self.flush_period = config.get(['west', 'data', 'flush_period'], self.default_flush_period)
-        self.iter_ref_h5_template = config.get(['west', 'data', 'data_refs', 'iteration'], None)
-        self.store_h5 = self.iter_ref_h5_template is not None
+
+        # Path to per-iter h5 file
+        self.iter_ref_h5_path_template = config.get(['west', 'data', 'data_refs', 'iteration'], None)
```
jpthompson17: One more nitpick: What does "ref" refer to in the name `iter_ref_h5_path_template`? I realize it's been in the name since the HDF5 framework was added, but elsewhere (in method names) we refer to `iter_h5file`. I'm wondering if `iter_h5file_path_template` might be clearer.

jpthompson17 (approved):
Looks good to me (aside from a couple nitpicks).

src/westpa/core/data_manager.py
jeremyleung521: `ref` is probably a reference to 'data_refs', the key from the yaml file.

Would `iter_h5_path_template` work a little better? Just so it's different enough from `iter_h5_template_file_path`?

src/westpa/core/trajectory.py
jpthompson17: Got it. Seems like there's good reason to ignore PEP 8 here.

src/westpa/core/data_manager.py
jpthompson17: `iter_h5_path_template` works (and is probably easier to distinguish from `iter_h5_template_file_path`).

PR #492, opened 2025-06-03
Extra Frames in per-iter HDF5 Files + seglog_loader Fix

jeremyleung521:
## Describe the changes made

In specific situations (like after restarting a simulation), the per-iter HDF5 file might contain extra frames for certain segments because the segment traj data might've already been written before the restart but the segment status itself did not update. After the segment is rerun after the restart, the new coordinates are also saved/appended into the per-iter H5, leading to extra frames.

The fix itself is a little messy because pytables/hdf5 API/specs does not allow one to remove rows in an `EArray`. If the existing number of rows is less than or equal the final needed number of rows (usually pcoord_len -1), extra rows are added as needed and the data for the old frames are rewritten with the new segment info. Other wise, extra rows' pointers are replaced with [-n_iter, seg_id] to distinguish from the newer frames.

This PR also fixes the broken seglog_loader and adds much needed tests for the HDF5 Framework

- [x] Do a check before writing segments into per-iter HDF5 file
- [x] new `seglog_writer` for recovering seglogs
- [x] Tests for per-iter HDF5 File (and restart/seglog loader/writer)
- [x] Fixes `Seglog_loader`

## Major files changed

- [x] src/westpa/core/h5io.py
- [x] tests/test_h5io.py
- [x] tests/conftest.py

## Status


Slimygen (approved):
Everything looks good.

PR #493, opened 2025-06-03
Analysis update

jpthompson17:
## Describe the changes made

This PR includes a few updates to the `westpa.analysis` subpackage:
- Add some unit tests for the `Run` API.
- Fix bugs in the `Run.__contains__` and `Iteration.num_bins` methods, and fix return type of `Run.num_iterations` (int instead of numpy int64).
- Ensure MDTraj is installed when initializing a `BasicMDTrajectory` or a `HDF5MDTrajectory` reader. This lays some groundwork for making MDTraj an optional dependency.

## Major files changed

- `src/westpa/analysis/core.py`
- `src/westpa/analysis/trajectories.py`
- `tests/test_analysis.py` (added)


jeremyleung521 (commented):
Was able to test this and was able to draw a directed graph (not sure about accuracy though, but the tests seem to do it). A few issue about this though:

a) No documentation about how to use it (once you have a Digraph, or for someone not familiar with networkx)
b) Graph itself is not labeled, so it's difficult to understand iterations/segment numbers (and it gets pretty complicated real quickly)
c) Graph is done on `Run` object, so if you only want to see the history graph N-iterations away you need to do some fancy truncating/deleting of your west.h5 file. 

Otherwise, great addition!

How I tested this:
```
import networkx as nx
import matplotlib.pyplot as plt
from westpa.analysis import Run

run = Run('west.h5')
nx.draw(run.history_graph())
plt.show()
```

jeremyleung521 (approved):

PR #534, opened 2025-12-01
Improve error message for NaN/inf values in w_pdist (fixes #170)

### LINKED ISSUE #170: w_pdist ValueError: Value out of bin boundaries
## This is part of an initiative to improve error messages so end users can better know how to debug their simulations.

To return to the hub, click on #163.

### Sample Error:
```
Traceback (most recent call last):
  ....
  File "/Users/russojd/anaconda3/envs/westpa-2020/westpa-2020.03/lib/west_tools/w_pdist.py", line 53, in _remote_bin_iter
    histnd(dset[:,ipt,:], binbounds, weights, out=iter_hist, binbound_check = False, ignore_out_of_range=ignore_out_of_range)
  File "fasthist/_fasthist.pyx", line 39, in fasthist._fasthist.histnd
  File "fasthist/_fasthist.pyx", line 101, in fasthist._fasthist.histnd
  File "fasthist/_fasthist.pyx", line 226, in fasthist._fasthist._histnd
ValueError: value nan at index 194 out of bin boundaries in dimension 0
```

### Relevant code:
`w_pdist`

### Potential causes:
Walker blowing up

This simulation was run for 3841 iterations. I managed to bracket in the problematic iteration to iteration 2752 by tweaking `--first-iter and --last-iter`. Here, row 191 of `west.h5/iterations/iter_00002752` shows that after 4 steps, this particular walker blew up and started reporting an infinity and then `NaN`s. This walker continued to report NaNs in subsequent iterations.

### Potential fixes:
This error should more explicitly indicate that a walker has blown up.

However, this does raise some additional questions -- `w_fluxanl` ran with no complaints on this data, and WESTPA apparently encountered these `NaN`s and silently kept running. Seems like we should at least raise a warning in the output of every iteration if there's a blown-up walker.

shubham-programmerr:
## Issue Number
Fixes #170

---

## Describe the changes made
Previously, `w_pdist` would fail with a low-level `ValueError` from `fasthist.histnd`
(e.g. "value nan at index … out of bin boundaries") when one or more walkers produced
NaN/inf values in the progress coordinate data. This error message was misleading and
did not clearly indicate that a walker had blown up.

This change adds an explicit check for non-finite (NaN/inf) values in `_remote_bin_iter`
before calling `histnd`. If such values are detected:
- an informative log message is emitted including the iteration number, time index,
  and affected segment indices,
- a clearer `ValueError` is raised explaining that one or more walkers have produced
  NaN/inf values and the simulation likely involves a blown-up walker.

---

- [x] Detect NaN/inf values in per-timepoint progress coordinate slices before histogramming
- [x] Provide clearer, user-friendly error message for blown-up walkers
- [ ] Add a targeted test for this case (future work, if maintainers prefer)

---

## Major files changed
- `west_tools/w_pdist.py`

---

## Status

---

## Additional context
I attempted to build and run the test suite locally, but installation fails on my system
(Python 3.13) with a `failed-wheel-build-for-install` error while building WESTPA.
I plan to retry with a supported Python version (e.g. 3.10/3.11). The change itself is
small and localized, and I’m happy to iterate based on CI feedback or maintainer guidance.

### DIFF
--- a/src/westpa/cli/tools/w_pdist.py
+++ b/src/westpa/cli/tools/w_pdist.py
@@ -68,6 +68,8 @@ def _remote_bin_iter(iiter, n_iter, dsspec, wt_dsspec, initpoint, binbounds, ign
     # normalize histogram
     normhistnd(iter_hist, binbounds)
     return iiter, n_iter, iter_hist
+
+
 class WPDist(WESTParallelTool):
     prog = 'w_pdist'
     description = '''\
--- a/src/westpa/cli/tools/w_pdist.py
+++ b/src/westpa/cli/tools/w_pdist.py
@@ -58,15 +58,50 @@ def _remote_bin_iter(iiter, n_iter, dsspec, wt_dsspec, initpoint, binbounds, ign
     npts = dset.shape[1]
     weights = wt_dsspec.get_iter_data(n_iter)
 
+    # Skip initial timepoint(s) if requested
     dset = dset[:, initpoint:, :]
+
     for ipt in range(npts - initpoint):
-        histnd(dset[:, ipt, :], binbounds, weights, out=iter_hist, binbound_check=False, ignore_out_of_range=ignore_out_of_range)
+        # Slice for this timepoint: shape (n_segments, n_dims)
+        slice_ = dset[:, ipt, :]
+
+        # 🔍 NEW: check for NaN / inf values before calling histnd
+        if not np.all(np.isfinite(slice_)):
+            # Find which segments are bad (optional, but helpful)
+            bad_mask = ~np.isfinite(slice_)
+            bad_seg_indices = np.where(bad_mask.any(axis=1))[0]
+
+            log.error(
+                "Detected non-finite (NaN/inf) progress coordinate values in iteration %d "
+                "for segment indices %s at time index %d. "
+                "This usually indicates that one or more walkers have blown up.",
+                n_iter,
+                bad_seg_indices.tolist(),
+                ipt + initpoint,
+            )
+
+            raise ValueError(
+                f"Non-finite (NaN/inf) progress coordinate values detected in iteration {n_iter} "
+                f"at time index {ipt + initpoint} for one or more walkers. "
+                "This likely indicates a blown-up walker; please inspect your simulation."
+            )
+
+        # Original histogram update
+        histnd(
+            slice_,
+            binbounds,
+            weights,
+            out=iter_hist,
+            binbound_check=False,
+            ignore_out_of_range=ignore_out_of_range,
+        )
 
     del weights, dset
 
     # normalize histogram
     normhistnd(iter_hist, binbounds)
     return iiter, n_iter, iter_hist
+
 
 
 class WPDist(WESTParallelTool):
--- a/src/westpa/cli/tools/w_pdist.py
+++ b/src/westpa/cli/tools/w_pdist.py
@@ -28,28 +28,6 @@ def isiterable(x):
         return True
 
 
-def _remote_min_max(ndim, dset_dtype, n_iter, dsspec):
-    try:
-        minval = np.finfo(dset_dtype).min
-        maxval = np.finfo(dset_dtype).max
-    except ValueError:
-        minval = np.iinfo(dset_dtype).min
-        maxval = np.iinfo(dset_dtype).max
-
-    data_range = [(maxval, minval) for _i in range(ndim)]
-
-    dset = dsspec.get_iter_data(n_iter)
-    for idim in range(ndim):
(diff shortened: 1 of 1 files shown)


shubham-programmerr:
Thanks for the review! CI is currently waiting for approval to run on this fork.
Happy to update anything based on feedback.

src/westpa/cli/tools/w_pdist.py
```diff
@@ -58,9 +58,43 @@ def _remote_bin_iter(iiter, n_iter, dsspec, wt_dsspec, initpoint, binbounds, ign
...
 
+    # Skip initial timepoint(s) if requested
     dset = dset[:, initpoint:, :]
+
     for ipt in range(npts - initpoint):
-        histnd(dset[:, ipt, :], binbounds, weights, out=iter_hist, binbound_check=False, ignore_out_of_range=ignore_out_of_range)
+        # Slice for this timepoint: shape (n_segments, n_dims)
+        slice_ = dset[:, ipt, :]
+
+        # 🔍 NEW: check for NaN / inf values before calling histnd
```
jeremyleung521: No emoji in the code please.

jeremyleung521 (commented):
Thanks for contributing. 

1. `_remote_bin_iter()` can be called in parallel, so you need to take into account that you might get a lot of duplicate errors/log messages in succession.
2. This is an analysis tool, and you're making a lot of assumptions regarding the simulation portion. As per #170, a check/error in the simulation portion/pcoord processing might be more apt.
3. If we are doing this, why are we complicating this and not directly checking/raising error in cython/ `histnd()`?

P.S. If you're having trouble with building the test suite, feel free to post the whole error message in a new issue. Python 3.13 IS supported.

shubham-programmerr:
Thanks a lot for the detailed feedback, Jeremy – this is really helpful.

1. You are right that `_remote_bin_iter()` can run in parallel, so doing
   per-timepoint checks and logging inside that loop could easily lead to a
   lot of duplicate messages. I hadn’t fully considered the parallel execution
   aspect.

2. I also see your point about this being an analysis tool and my change
   effectively making assumptions about the simulation/pcoord side. Moving the
   check closer to where the progress coordinates are produced would indeed be
   cleaner.

3. For this PR I was initially aiming for a small, user-facing improvement in
   the error message, not changing the core behaviour. Based on your comment,
   I can instead simplify this to only wrapping the `ValueError` from
   `histnd()` with additional context (iteration and time index), without
   adding any extra logging or simulation assumptions. If you’d prefer, I’m
   also happy to explore adding the check/error inside the Cython
   `histnd()`/`fasthist` layer instead, and would appreciate pointers to the
   best place to do that.

Regarding tests: thanks for clarifying that Python 3.13 is supported. I was
seeing a build failure when trying `pip install -e .` on my machine. I’ll try
again in a clean virtual environment and, if I still hit the issue, I’ll open
a separate issue with the full error log as you suggested.

shubham-programmerr:
Thank you for the detailed feedback!

I understand the concerns:

1) _remote_bin_iter is parallelized and duplicate logging is undesirable.
2) You are right that this may be better handled in the simulation / pcoord stage.
3) Raising at histnd() level makes more architectural sense.

Before making changes, I’d like guidance on the preferred direction:

Would you prefer:
(A) Silently skipping non-finite values with a single warning per iteration?
(B) Moving detection closer to the propagator / pcoord stage?
(C) Handling the check directly in fasthist / histnd Cython layer?
(D) Remove this approach and instead propose warning + documentation only?

Once I know the preferred approach, I will update the PR accordingly.

shubham-programmerr:
Thanks for the review!

I’ve updated the implementation as discussed (now just wrapping the
`histnd` `ValueError` with extra context, without extra logging or NaN
checks) and I’m happy to adjust further based on any additional feedback.

jeremyleung521:
* Your code will catch false positives `ValueError` from `histnd()`, e.g. raised when the binbounds are not monotonically increasing etc.
* You're failing lint. Please check the  [developer's guide](https://github.com/westpa/westpa/wiki/WESTPA-Developer%27s-Guide) and [this other page](https://github.com/westpa/westpa/wiki/Setting-Up-a-Development-Environment) for getting pre-commit working.
* Why did you delete the `_remote_min_max()` function?
* Please get a working environment (and run the tests locally) before reopening this PR.

shubham-programmerr:
Thanks for the detailed feedback and for taking the time to review.

I understand that my current approach isn’t the right fit and that my local
environment was not properly set up (tests + pre-commit). I’ll take a step
back, set up a correct environment, and focus on smaller issues first before
attempting this one again.

Thanks again for the guidance.

Issue #533, opened 2025-11-30
w_run hangs indefinitely in `ProcessesWorkManager.shutdown()` after PyTables NodeError when writing `/log/<iter_seg>`


HScheiber:
## Prerequisites
* [ ] Can you reproduce the problem in safe mode?
* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

### Environment

- WESTPA: 2022.12
- Python: 3.12.3
- OS: Linux (AWS Batch managed EC2, Amazon Linux–like)
- Work manager: processes
- h5py version: 3.15.1

### Summary

When the HDF5 data (specifically, /log entries for a given iteration/segment) is corrupted and w_run hits a PyTables NodeError such as:

`tables.exceptions.NodeError: group /log/188_841 already has a child node named "data"`

`w_run` prints the `exception caught; shutting down` error and attempts to clean up, but never exits. The main thread remains blocked in `ProcessesWorkManager.shutdown()` inside `_empty_queues`, so the process does not terminate. I noticed because my outer driver never regains control of the thread. It just hangs indefinitely waiting for `w_run` to complete.

The core problem is: after an exception in `DataManager.update_iter_h5file`, the processes work manager’s shutdown/queue-drain logic can hang indefinitely, so `w_run` does not terminate.

### Symptoms / Observed behavior

From the w_run output:

```shell
Waiting for segments to complete...
exception caught; shutting down
-- ERROR    [w_run] -- error message: group ``/log/188_841`` already has a child node named ``data``
-- ERROR    [w_run] -- Traceback (most recent call last):
  File ".../westpa/cli/core/w_run.py", line 61, in run_simulation
    sim_manager.run()
  File ".../westpa/core/sim_manager.py", line 767, in run
    self.propagate()
  File ".../westpa/core/binning/acevt_manager.py", line 72, in propagate
    self.data_manager.update_segments(self.n_iter, incoming)
  File ".../westpa/core/data_manager.py", line 1128, in update_segments
    self.update_iter_h5file(n_iter, segments)
  File ".../westpa/core/data_manager.py", line 619, in update_iter_h5file
    outf.write_segment(segment, True)
  File ".../westpa/core/h5io.py", line 729, in write_segment
    self._create_array(
  File ".../tables/file.py", line 1230, in create_array
  ...
tables.exceptions.NodeError: group ``/log/188_841`` already has a child node named ``data``
```

[...]

PR #537, opened 2025-12-01
Fix Processes work manager silent hang

jeremyleung521:
## Issue Number

#533 

## Describe the changes made

Effectively I reorganized the processes manager so during shutdown:
1. Clean up the queues
2. Throw `shutdown_sentinels` to shutdown task_queue
3. throw `shutdown_sentinel` to shutdown result_queue (also modified it so it is no longer blocking, with the default timeout)
4. Send shutdown to the queue retrieval loops via the "shutdown_received" Event
5. Clean queues again
6. Shutdown all workers

- Also fixed the bug where logs (for HDF5 Framework) were not deleted properly.
- Also added functionality for sim_manager to detect whether the work_manager is running (and subsequently shutdown the future-related `self.work_manager.wait_any()`
- safe_extract will use the 'fully_trusted' filter when tarring/untarring. This will unify the behavior between python <3.14 and python >3.14
https://docs.python.org/3/library/tarfile.html#extraction-filters

- [x] Fix error where existing log dataset leads to hang processes
- [x] Update processes work manager to be more robust

## Major files changed

- [x] src/westpa/core/h5io.py
- [x] src/westpa/work_managers/processes.py

## Status

### DIFF
--- a/src/westpa/core/h5io.py
+++ b/src/westpa/core/h5io.py
@@ -723,8 +723,8 @@ class WESTIterationFile(HDF5TrajectoryFile):
             )
 
         if slog is not None:
-            if self._has_node('/log', str(segment.seg_id)):
-                self._remove_node('/log', name=str(segment.seg_id), recursive=True)
+            if self._has_node('/log', name='%d_%d' % (segment.n_iter, segment.seg_id)):
+                self._remove_node('/log', name='%d_%d' % (segment.n_iter, segment.seg_id), recursive=True)
 
             self._create_array(
                 '/log/%d_%d' % (segment.n_iter, segment.seg_id),
--- a/src/westpa/core/h5io.py
+++ b/src/westpa/core/h5io.py
@@ -117,7 +117,7 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
         if not is_within_directory(path, member_path):
             raise Exception("Attempted Path Traversal in Tar File")
 
-    tar.extractall(path, members, numeric_owner=numeric_owner, filter='data')
+    tar.extractall(path, members, numeric_owner=numeric_owner, filter='fully_trusted')
 
 
 #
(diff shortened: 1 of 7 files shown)


src/westpa/core/h5io.py
```diff
@@ -117,7 +117,7 @@ def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
         if not is_within_directory(path, member_path):
             raise Exception("Attempted Path Traversal in Tar File")
 
-    tar.extractall(path, members, numeric_owner=numeric_owner)
+    tar.extractall(path, members, numeric_owner=numeric_owner, filter='data')
```
HScheiber: This change causes parent trajectories not to be passed to children when saving restart files (e.g. *.trr files) using the HDF5 framework. It breaks simulations.

jeremyleung521: Thanks for catching this! Python 3.14 defaults to the 'data' filter so I thought we'd unify it too, but I'll swap it right back to 'fully_trusted' and not bother with it (which is default for python <3.14).

src/westpa/work_managers/processes.py
```diff
@@ -109,6 +109,7 @@ def results_loop(self):
                     raise AssertionError('unknown message {!r}'.format((message, task_id, payload)))
 
         log.debug('exiting results_loop')
+        return
```
jpthompson17, line 112: Why the explicit `return`?

jpthompson17 (approved):

src/westpa/work_managers/processes.py
jeremyleung521: Mostly for code parity between `results_loop` and `task_loop`. An explicit `return` returns None, which effectively works the same as not having the line at all.

Bailyford (approved):
Looks good.

PR #540, opened 2025-12-31
Dask-based work manager

jpthompson17:
This PR adds a new work manager built on the [dask.distributed](https://github.com/dask/distributed) library.

Like the MPI work manager, the Dask work manager is not available by default. To enable it, install WESTPA with the `dask` extra:
```sh
$ python -m pip install westpa[dask]
```

## Usage

Run on the default local cluster:

```sh
$ w_run --work-manager dask
```

Run on a local cluster with 6 workers:
```sh
$ w_run --work_manager dask --n-workers 6
```

Start up a cluster and point WESTPA to the scheduler address:
```sh
$ dask scheduler
INFO -   Scheduler at:     tcp://10.0.0.164:8786
$ dask worker 10.0.0.164:8786
$ w_run --work-manager dask --dask-scheduler-address 10.0.0.164:8786
```

Start up a cluster and point WESTPA to the scheduler file:
```sh
$ dask scheduler --scheduler-file /path/to/scheduler.json
$ dask worker --scheduler-file /path/to/scheduler.json
$ w_run --work-manager dask --dask-scheduler-file /path/to/scheduler.json
```

## Status


pyproject.toml
```diff
@@ -47,6 +47,8 @@ dependencies = [
   "tqdm",
   "pandas",
   "tables",
+  "dask",
```
jeremyleung521: Are we adding dask as a hard dependency? Or should we make it optional like mpi?

src/westpa/work_managers/dask.py
```diff
@@ -0,0 +1,136 @@
...
+from itertools import islice
+
+import dask.distributed as distributed
+
+import westpa
+import westpa.work_managers as work_managers
+from .core import WorkManager
+
+
+class _DaskFutureWrapper:
```
jeremyleung521, line 14: If it's WMFuture-like, why isn't it a subclass of that?

tests/test_work_managers/test_dask.py
```diff
@@ -0,0 +1,48 @@
...
+
+NUM_TASKS = 5
+QUEUE_SIZE = 2
+
+
+@pytest.fixture(scope='module')
+def work_manager():
+    return DaskWorkManager()
+
+
```
jeremyleung521, line 21: I would put all the tests into [a single class](https://github.com/westpa/westpa/blob/westpa2/tests/test_work_managers/test_processes.py#L20), just so it's easier to track. Would be nice to add [some timeouts ](https://github.com/jeremyleung521/westpa/blob/d8b236854f828b135a2a7d13411a4eaf06085cc1/tests/test_work_managers/test_processes.py#L32)(since work-manager related things are very flaky on GH Action runners)


```suggestion
@pytest.mark.timeout(5)
```

jeremyleung521 (commented):
Add `dask` and `dask-distributed` to these files for CI. We use `--no-deps` with pip to speed up Runner environment setup time, so those packages were never installed.

https://github.com/jpthompson17/westpa/tree/dask-work-manager/devtools/conda-envs

Just a quick scan. Will be back later after I actually test out the code.

pyproject.toml
jpthompson17: My sense was we were adding it as a hard dependency, but we could make it optional. Do you have any reservations about making it a hard dependency?

src/westpa/work_managers/dask.py
jpthompson17: I initially wrote it as a subclass, but we don't need all the `WMFuture` attributes (e.g., `_condition`, `_exception`, `_watchers`, even `task_id`) since we're calling through to Dask. My feeling was that if we weren't going to call `super().__init__()`, it made more sense to use [structural subtyping](https://typing.python.org/en/latest/reference/protocols.html) as opposed to nominal subtyping. (Perhaps we could be more explicit about this by defining `WMFuture` as a `Protocol`.)

pyproject.toml
jpthompson17: Added dask and distributed to CI environment: https://github.com/westpa/westpa/pull/540/commits/c0087ba03ff6320071d159e7023b267555933e31

jeremyleung521:
On some of the configs, all the tests pass, but pytest refuses to completely shutdown and thus stalls. This suggests it's either a runner issue, or the dask tests weren't shutting down/cleaning up properly.

pyproject.toml
jeremyleung521: Just (in general) trying to limit the number of dependencies. Got burned too much by badly maintained dependencies. :) Especially since it's optional (i.e. processes, serial and threads wm work fine without dask), it's good to cordon them off.

src/westpa/work_managers/dask.py
```diff
@@ -0,0 +1,248 @@
...
+        kwargs = {'client': {}, 'cluster': {}}
+        kwargs['client']['address'] = wmenv.get_val('dask_scheduler_address')
+        kwargs['client']['scheduler_file'] = wmenv.get_val('dask_scheduler_file')
+
+        n_workers_val = wmenv.get_val('n_workers')
+        kwargs['n_workers'] = int(n_workers_val) if n_workers_val is not None else None
+        n_threads_val = wmenv.get_val('dask_n_threads_per_worker')
+        kwargs['threads_per_worker'] = int(n_threads_val) if n_threads_val is not None else None
+        memory_limit_val = wmenv.get_val('dask_memory_limit')
+        kwargs['client']['memory_limit'] = memory_limit_val if memory_limit_val is not None else 'auto'
```
jeremyleung521:
```suggestion
        kwargs['cluster']['memory_limit'] = memory_limit_val if memory_limit_val is not None else 'auto'
```

src/westpa/work_managers/dask.py
```diff
@@ -0,0 +1,248 @@
...
+        Usually called by each CLI tool."""
+        if not self.running:
+            if isinstance(self.client, distributed.Client):
+                # distributed.Client is already initialized and given by user
+                self._local_cluster = self.client.cluster
+            elif isinstance(self.client, dict):
+                # Start client/cluster based on user-supplied arguments for distributed.client
+                self.client = distributed.Client(
+                    n_workers=self.supplied_n_workers, threads_per_worker=self.supplied_n_threads, **self.client
+                )
```
jeremyleung521:
```suggestion
                if self.client['address'] is not None or self.client['scheduler_file'] is not None:
                    self.client = distributed.Client(**self.client)
                else:
                    self.client = distributed.Client(
                        n_workers=self.supplied_n_workers, threads_per_worker=self.supplied_n_threads, **self.client, **self.startup_kwargs
                    )
                self._local_cluster = self.client.cluster
```

src/westpa/work_managers/dask.py
```diff
@@ -0,0 +1,248 @@
...
+        group.add_argument(
+            wmenv.arg_flag('dask_n_threads_per_worker'),
+            metavar='THREADS_PER_DASK_WORKER',
+            type=int,
+            help="Number of threads per dask worker.",
+        )
+        group.add_argument(
+            wmenv.arg_flag('dask_memory_limit'),
+            metavar='MEMORY_LIMIT',
+            type=int,
```
jeremyleung521:
```suggestion
            type=str,
```

jeremyleung521 (changes requested):
I messed up so it was impossible to pass in a json or an address of an existing client/cluster. With the following three suggested changes, it should work.

I also noticed that currently, regardless of how the cluster/client is started, shutting down w_run will shut down the cluster/clients. I don't know if that should be the intended behavior.

Otherwise, everything is working and I approve :)

src/westpa/work_managers/dask.py
jpthompson17: Thanks :) I'll take a look at the shutdown issue. I'd prefer for the cluster to stay online when w_run shuts down. What are your thoughts on this?

jeremyleung521: > Thanks :) I'll take a look at the shutdown issue. I'd prefer for the cluster to stay online when w_run shuts down. What are your thoughts on this?

I prefer that too, especially if users are providing the cluster/client. Otherwise it should shutdown completely.

Probably just need to set a flag in [startup](https://github.com/jpthompson17/westpa/blob/e283e905f124675022b51723aee06e8cafd60caa/src/westpa/work_managers/dask.py#L126) and depending on the flag, not go through with certain parts of the `shutdown` (https://github.com/jpthompson17/westpa/blob/e283e905f124675022b51723aee06e8cafd60caa/src/westpa/work_managers/dask.py#L152)

Might be worth adding a `force=True` thing to `shutdown` that bypasses the check too.

jpthompson17:
I made a few changes:
- Revise `shutdown()` method: Only close the cluster or client if it was created by the work manager.
- Remove `n_workers` and `threads_per_worker` parameters. All `LocalCluster` parameters now go in `kwargs`.
- Remove unused `map()` method.
- Don't pass `None` to `type_` in `WMEnvironment.get_val()`. This lets us use `type_` for optional arguments (similar to `type` in `argparse`).

jeremyleung521 (commented):
Approved, after https://github.com/jpthompson17/westpa/pull/4 is merged in.

jeremyleung521 (dismissed):

jpthompson17:
Thanks for https://github.com/jpthompson17/westpa/pull/4. I'm making a few more small changes.

jeremyleung521:
Everything's working well! Just one last thing though. I tried to run the following, but forgot to run the `dask worker` line.

> $ dask scheduler
> $ dask worker 10.0.0.164:8786
> $ w_run --work-manager dask --dask-scheduler-address 10.0.0.164:8786

west.log just silently hangs right before the `Waiting for segments to complete...` line. Then I realized my mistake and ran `dask worker ...`, where the simulation proceeds normally.

Is there a place to throw a warning saying no workers, after, say a 1 minute wait? Probably something from the scheduler in a `no-worker`/ `unrunnable` state (https://distributed.dask.org/en/latest/scheduling-state.html#distributed.scheduler.Scheduler)?

jeremyleung521:
> Everything's working well! Just one last thing though. I tried to run the following, but forgot to run the `dask worker` line.
> 
> > $ dask scheduler
> > $ dask worker 10.0.0.164:8786
> > $ w_run --work-manager dask --dask-scheduler-address 10.0.0.164:8786
> 
> west.log just silently hangs right before the `Waiting for segments to complete...` line. Then I realized my mistake and ran `dask worker ...`, where the simulation proceeds normally.
> 
> Is there a place to throw a warning saying no workers, after, say a 1 minute wait? Probably something from the scheduler in a `no-worker`/ `unrunnable` state (https://distributed.dask.org/en/latest/scheduling-state.html#distributed.scheduler.Scheduler)?

For this I think we can just throw a log.warning() at startup.

----
Also realized that we don't have a solution for this either here, for multi-gpu setups. `WM_PROCESS_INDEX` is not set, for obvious reasons, and we need an equivalent to expose the relevant CUDA_VISIBLE_DEVICES. I need to test this a little more on GPUs, so rescinding my approval for a bit..
https://github.com/jpthompson17/westpa-api-workshop/discussions/9

This is very analog, lol.
https://docs.dask.org/en/stable/gpu.html#specifying-gpus-per-machine

jpthompson17:
Starting a worker with
```
CUDA_VISIBLE_DEVICES=0 dask worker ...
```
exports `CUDA_VISIBLE_DEVICES=0` to the worker's environment. Do we need to pass more information than this?

jeremyleung521:
> Starting a worker with
> 
> ```
> CUDA_VISIBLE_DEVICES=0 dask worker ...
> ```
> 
> exports `CUDA_VISIBLE_DEVICES=0` to the worker's environment. Do we need to pass more information than this?

In terms of exposing the correct GPU to a program, nope. But there isn't a way to do this via the Python API as far as I can tell, which is the weirdest design choice ever.

jeremyleung521:
Getting this bug if I try to initialize the tutorial 7.1 simulation with the dask work manager... Digging deeper right now... This is after pulling all the newest commits from `develop`, btw.
```
# Set pointer to bstate and tstate
BSTATE_ARGS="--bstate-file $WEST_SIM_ROOT/bstates/bstates.txt"
TSTATE_ARGS="--tstate-file $WEST_SIM_ROOT/tstate.file"

# Run w_init
w_init \
  $BSTATE_ARGS \
  $TSTATE_ARGS \
  --segs-per-state 5 \
  --work-manager=dask "$@"
```

```
(westpa-dask) jml230@system76:~/dask-test/westpa_tutorials/tutorial7.1-basic-nacl$ ./init.sh
Updating system with the options from the configuration file
Creating HDF5 file '/home/jml230/dask-test/westpa_tutorials/tutorial7.1-basic-nacl/west.h5'
1 target state(s) present
Calculating progress coordinate values for basis states.
Updating system with the options from the configuration file
Traceback (most recent call last):
  File "/home/jml230/apps/miniforge3/envs/westpa-dask/bin/w_init", line 6, in <module>
    sys.exit(entry_point())
             ~~~~~~~~~~~^^
  File "/home/jml230/git/westpa-dask/src/westpa/cli/core/w_init.py", line 110, in entry_point
    initialize(
    ~~~~~~~~~~^
        args.tstates,
        ^^^^^^^^^^^^^
    ...<6 lines>...
        args.shotgun,
        ^^^^^^^^^^^^^
    )
    ^
  File "/home/jml230/git/westpa-dask/src/westpa/cli/core/w_init.py", line 217, in initialize
    sim_manager.initialize_simulation(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        basis_states=basis_states,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        suppress_we=shotgun,
        ^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jml230/git/westpa-dask/src/westpa/core/sim_manager.py", line 299, in initialize_simulation
    self.get_bstate_pcoords(basis_states)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
  File "/home/jml230/git/westpa-dask/src/westpa/core/sim_manager.py", line 240, in get_bstate_pcoords
    basis_states[fmap[future]] = future.get_result()
                 ~~~~^^^^^^^^
KeyError: _DaskFutureWrapper(None)
```

jeremyleung521:
The problem was that the `DaskFutureWrapper` consumes the Dask Future object (`self.future`) after `get_results()`, so it is unable to find the bstate id after `get_result()` in `get_bstate_pcoords()` sim manager.
https://github.com/jpthompson17/westpa/blob/4ee6e9710a1d2fa79a3ed5639057c43b50dc1f76/src/westpa/work_managers/dask.py#L42-L43

https://github.com/jpthompson17/westpa/blob/4ee6e9710a1d2fa79a3ed5639057c43b50dc1f76/src/westpa/core/sim_manager.py#L240

The quick, naive way around it is to find/set the bstate id first before consuming the future.

Another way is to mark each Dask Future based on the unique key (already generated by Dask), and have DaskWorkManager deal with the bookeeping/translation. This is better in the sense that the DaskWorkManager will can deal with removing stale (Dask) futures without having to worry about WESTPA's other parts using the Dask Futures as keys, causing lingering memory issues, but might require the DaskWorkManager to keep a "master" dictionary mapping keys to actual dask future objects.

jpthompson17:
When lines [42-43](https://github.com/jpthompson17/westpa/blob/4ee6e9710a1d2fa79a3ed5639057c43b50dc1f76/src/westpa/work_managers/dask.py#L42-L43) were added, why was the choice made to delete the reference to the Dask future? I.e.,
```py
if not discard:
    self._result = result
else:
    self.future = None
```
The [analog](https://github.com/westpa/westpa/blob/71243e390f4e9ce3e2fb80cb3268cddfcfb552bd/src/westpa/work_managers/core.py#L346) in `WMFuture` only deletes the `_result` attribute:
```py
if discard:
    del self._result
```

I'm wondering if we could change L40-43 to:
```py
if not discard:
    self._result = result
else:
    self._result = None
```
and change the implementation of `_DaskFutureWrapper.is_done()` back to:
```py
def is_done(self):
    return self.future.done()
```

jeremyleung521:
The reference was holding the (Dask) future around, preventing the scheduler from shutting down.

jeremyleung521:
The documentation implies that we cannot just remove dask futures results (leaving the future around). In most cases, only by removing reference to the future do we remove the remote data.

https://distributed.dask.org/en/latest/memory.html#clearing-data

jpthompson17:
Got it. What about caching the hash value, so the `_DaskFutureWrapper` remains hashable after the Dask future is ejected? E.g.
```py
    def __init__(self, future):
        self.future = future
        self._result = None
        self._hashval = hash(future)

    def __hash__(self):
        return self._hashval
```

jeremyleung521:
> Got it. What about caching the hash value, so the `_DaskFutureWrapper` remains hashable after the Dask future is ejected? E.g.
> 
> ```python
>     def __init__(self, future):
>         self.future = future
>         self._result = None
>         self._hashval = hash(future)
> 
>     def __hash__(self):
>         return self._hashval
> ```

That sounds fine except it won't help with this line, because `fmap`, made there is created based on the future object itself, not the hash:

https://github.com/westpa/westpa/blob/westpa2/src/westpa/core/sim_manager.py#L240

Changing this will require us to change the logic for all of the work managers... which seems... a little overkill.

jeremyleung521:
I'm ok with merging this after that final PR https://github.com/jpthompson17/westpa/pull/7 is merged. What an effort, but hopefully it's robust enough... we'll see!

PR #544, opened 2026-01-05
Fix Hang in Python 3.14 GH Actions

jeremyleung521:
## Describe the changes made

https://github.com/westpa/westpa/actions/runs/20641223356

All Python 3.14 tests began to hang on the `TestInstantiations/test_zeroMQ` test from `test_work_managers/test_environment.py`. This is due to the fact that the zeromq work manager is not properly handling the `ValueError` raised by `numpy 2.4.0` for inplace numpy resizing.

https://github.com/numpy/numpy/pull/30282

Since inplace resizing will (seemingly) be removed in the future anyways, we might as well just forgo that completely. This will increase memory usage, but that's what numpy wants.

I think overall we need to fix zeromq so the futures will handle the errors more elegantly. You can throw in a `raise ValueError`  in the [add_timer() method in src/westpa/work_managers/zeromq/core.py](https://github.com/westpa/westpa/blob/westpa2/src/westpa/work_managers/zeromq/core.py#L208) and it will hang similarly in any python version (tried it on Python 3.13, both numpy 2.3.5, 2.4.0). That'll require some more digging so I'll leave it to a different PR.

Hang can be reproduced via the following on python 3.14/numpy 2.4.0:
```
pytest -v tests/test_work_managers/test_environment.py -k 'testZeroMQ' 
```

- [x] Make Python 3.14 tests run

## Major files changed

- [x] src/westpa/work_managers/zeromq/core.py

## Status

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - "westpa2"
       - "develop"
+      - "py314-action-fix"
   pull_request:
     branches:
       - "westpa2"
@@ -39,15 +40,8 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest", "ubuntu-24.04-arm", "macos-15-intel", "macos-latest"]  # macos-15-intel is x86-64, macos-latest is arm64
-        python-version: ["3.11", "3.12", "3.13", "3.14"]
-        numpy-version: ["1", "2"]
-        exclude:
-          - python-version: "3.13"
-            numpy-version: "1"
-          - python-version: "3.14"
-            numpy-version: "1"
-          - os: "ubuntu-24.04-arm"
-            numpy-version: "1"
+        python-version: ["3.14"]
+        numpy-version: ["2"]
     steps:
       - uses: actions/checkout@v6
         with: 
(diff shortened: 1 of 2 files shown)


jeremyleung521:
Just saw this. Might be problematic. Will think a little bit about this. https://numpy.org/doc/stable/reference/generated/numpy.resize.html

> If the new array is larger than the original array, then the new array is filled with repeated copies of a. Note that this behavior is different from a.resize(new_shape) which fills with zeros instead of repeated copies of a.

jeremyleung521:
> Just saw this. Might be problematic. Will think a little bit about this. https://numpy.org/doc/stable/reference/generated/numpy.resize.html
> 
> > If the new array is larger than the original array, then the new array is filled with repeated copies of a. Note that this behavior is different from a.resize(new_shape) which fills with zeros instead of repeated copies of a.

The bigger picture is that we're adding an extra position to the arrays and reassigning it to a desired value. Not sure why we're using numpy arrays here (lists would be more than sufficient). Anyhow, just so everything is consistent and we prevent race conditions, I'll force a copy and will do the extending + addition in a single step (via `numpy.pad`).

Bailyford (approved):
Looks good.

Issue #545, opened 2026-01-08
New warnings from old binning pickles


jeremyleung521:
## Prerequisites
* [x] Can you reproduce the problem in safe mode?
* [x] Are you running the latest version?
* [x] Are you reporting to the correct repository?
* [x] Did you perform a cursory search?

## Describe the bug

```
tests/test_analysis.py::test_sink
tests/test_analysis.py::test_binning
  /home/runner/miniconda3/envs/test_env/lib/python3.13/site-packages/westpa/tools/binning.py:82: VisibleDeprecationWarning: dtype(): align should be passed as Python or NumPy boolean but got `align=0`. Did you mean to pass a tuple to create a subarray type? (Deprecated NumPy 2.4)
    mapper = pickle.loads(pkldat)

tests/test_tools/test_w_multi_west.py::Test_W_Multi_West::test_run_w_multi_west_noaux
  /home/runner/miniconda3/envs/test_env/lib/python3.13/site-packages/westpa/cli/tools/w_multi_west.py:69: VisibleDeprecationWarning: dtype(): align should be passed as Python or NumPy boolean but got `align=0`. Did you mean to pass a tuple to create a subarray type? (Deprecated NumPy 2.4)
    mapper = pickle.loads(pkldat)
```

https://numpy.org/devdocs/reference/generated/numpy.dtype.html#numpy-dtype
https://github.com/westpa/westpa/actions/runs/20825172816/job/59824429583

Result of this PR: https://github.com/numpy/numpy/pull/29301

From reading https://github.com/numpy/numpy/issues/8158, the solution is probably to remake the pickles because apparently is an old bug in the `pickle` module to save align as 0/1 instead of True/False.

## To Reproduce

1. Run one of the tests (test_binning/test_sink/test_run_w_multi_west_noaux) with numpy >= 2.4.0

## Expected behavior

No Warning

## Actual behavior

See description

## WESTPA version and platform

numpy 2.4.0, any WESTPA version

## Additional context

Probably requires digging through the pickled bins to figure out why align is set to 0 (instead of False). Nothing really stands out by searching for dtype in the WESTPA code.

jeremyleung521:
Fixed in #562.

PR #548, opened 2026-01-16
Repickle BinMappers and Refactor westpa.analysis tests

jeremyleung521:
## Issue Number

#545

## Describe the changes made

Fixes it so the binmapper pickles objects in `west_ref.h5` do not raise warnings.
While fixing that, I refactored the westpa.analysis tests.

- [x] Make sure test files are forward compatible with future numpy versions

## Major files changed

- [x] tests/

## Status


jpthompson17 (commented):
To my eyes, the proposed test refactoring accomplishes consistency at the expense of clarity. It adds a redundant layer of nesting (`test_analysis.TestAnalysis`) and makes the value/behavior of the test fixture less easy to grasp at a glance. Instead of finding the fixture in the same module as the functions that use it, now we have to go to `conftest.py` to find it. Unlike the `run` fixture, which evaluated directly to a `Run` object, the `ref_analysis` fixture operates in an indirect way: it attaches a `Run` object to the enclosing test class. If we're new to the code, it's only by reading this far that we can know for sure where the `TestAnalysis.run` attribute came from, and what its type is.

I get wanting things to be consistent, but is there a clear need to use the pattern in `conftest.py` across the entire test suite?

jeremyleung521:
> To my eyes, the proposed test refactoring accomplishes consistency at the expense of clarity. It adds a redundant layer of nesting (`test_analysis.TestAnalysis`) and makes the value/behavior of the test fixture less easy to grasp at a glance. Instead of finding the fixture in the same module as the functions that use it, now we have to go to `conftest.py` to find it. Unlike the `run` fixture, which evaluated directly to a `Run` object, the `ref_analysis` fixture operates in an indirect way: it attaches a `Run` object to the enclosing test class. If we're new to the code, it's only by reading this far that we can know for sure where the `TestAnalysis.run` attribute came from, and what its type is.
> 
> I get wanting things to be consistent, but is there a clear need to use the pattern in `conftest.py` across the entire test suite?

It's what we decided on back in 2021:
https://github.com/westpa/westpa/wiki/Writing-pytests-for-WESTPA

Also what is recommended in pytest's documentation for shared fixtures: 
https://docs.pytest.org/en/stable/reference/fixtures.html#conftest-py-sharing-fixtures-across-multiple-files

Also grouping tests in classes it'll be much easier for me to schedule them when I finally get around to writing a scheduler for parallel test execution with `pdist-xdist`, where tests could run out of order and on different processes.

The CI error is from the fact that I remade the pickles in a newer numpy version... Will repickle them in numpy 1.25/python 3.11, i guess.

jpthompson17 (approved):

PR #460, opened 2024-11-12
RNG Modernization

jeremyleung521:
## Describe the changes made

This PR isolate the RNG from the global `random` and/or `np.random` modules into independent `numpy.random.Generator` (see [here](https://numpy.org/doc/stable/reference/random/index.html)), protecting the key portions of WESTPA from other programs that might seed the RNG irresponsibly. For the sake of consistency, we will continue to use the MT19937 algorithm.

The RNG in Cythonized parts are not replaced because an independent `numpy.random.Generator` requires locking the GIL every time you request a number, and that's just not efficient.

The fasthist testing script `/src/westpa/fasthist/__main__.py` has been refactored into `test_fasthist.py`, to show/test that it's faster than `numpy.histogramdd`.

## Goals and Outstanding Issues

- [x] Modernize and isolate WESTPA RNG
- [x] Refactor /src/westpa/fasthist/__main__.py into test_fasthist.py

## Major files changed

- [x] src/westpa/core/propagators/executable.py
- [x] src/westpa/core/we_driver.py
- [x] and other tests and old tools (see below)

## Status

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] I have read the CONTRIBUTING document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #567, opened 2026-02-27
Initialize RNG a little later in the propagator

jeremyleung521:
## Issue Number

#460 

## Describe the changes made

An initialized RNG might be copied verbatim within the PropagatorExecutable if using the `processes` work manager. This is a result of the `os.fork()` on UNIX.

The bug can be reproduced when running tutorial 7.2 as is. All the trajectories will have the same seed. I can reproduce this starting from 2022.12. Seemingly works fine on 2022.10 (but I'm having trouble with holding all the variables constant--dependency version, python version etc--due to how old it is) so I'm semi-sure it's due to #460.

- [x] Fix the RNG problem with certain work managers (`processes`)

## Major files changed

## Status

- [x] I have read the AI POLICY document.
- [x] This PR conforms to WESTPA's AI policy.
- [x] I have declared all usage of AI in the PR description.

### DIFF
--- a/src/westpa/core/propagators/executable.py
+++ b/src/westpa/core/propagators/executable.py
@@ -221,6 +221,8 @@ class ExecutablePropagator(WESTPropagator):
         as a dictionary, suitable for use in ``os.environ.update()`` or as the ``env`` argument to
         ``subprocess.Popen()``. Every child process executed by ``exec_child()`` gets these.'''
 
+        # Initialize rng here when we need our first number. This will prevent
+        # certain work managers (e.g., `processes`) from reusing an initialized RNG.
         if self.rng is None:
             self.rng = Generator(MT19937())
 
--- a/src/westpa/core/propagators/executable.py
+++ b/src/westpa/core/propagators/executable.py
@@ -100,8 +100,8 @@ class ExecutablePropagator(WESTPropagator):
         self.initial_state_ref_template = config['west', 'data', 'data_refs', 'initial_state']
         store_h5 = config.get(['west', 'data', 'data_refs', 'iteration']) is not None
 
-        # Create a persistent RNG for each worker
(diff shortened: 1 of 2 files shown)


src/westpa/core/propagators/executable.py
```diff
@@ -221,6 +221,11 @@ def random_val_env_vars(self):
         as a dictionary, suitable for use in ``os.environ.update()`` or as the ``env`` argument to
         ``subprocess.Popen()``. Every child process executed by ``exec_child()`` gets these.'''
 
+        # Initialize rng here when we need our first number. This will prevent certain work managers
+        # (e.g., `processes`) from reusing an RNG that might be initialized too early (i.e. before forking).
+        if self.rng is None:
+            self.rng = Generator(MT19937())
```
jpthompson17, line 227: The [recommended](https://numpy.org/devdocs/reference/random/generator.html#numpy.random.Generator) Generator constructor is `default_rng()`. Honestly, though, that's a little confusing when using a non-default BitGenerator like MT19937, so I'm neutral on whether we follow NumPy's recommendation here.

jpthompson17 (approved):
This looks okay to me.

src/westpa/core/propagators/executable.py
jeremyleung521: Yea. when I first switched from `random.random` global generator to `np.random.Generator`, I thought about switching to PCG64 (which is what `default_rng` is ). But then I saw how they had to develop [PCG64DXSM ](https://numpy.org/devdocs/reference/random/bit_generators/pcg64dxsm.html) to counter some of the problems with that and figured it's probably safer to use `MT19937` (which was what we've/built-in `random` been using), something more tried and true (and much older than PCG64)...

`Generator(MT19937())` is definitely clearer.

Issue #543, opened 2026-01-02
Add w_reverse CLI tool


jpthompson17:
Add a new CLI tool, based on @darianyang's [prototype](https://github.com/darianyang/misc_we/blob/main/w_reverse.py), to assist with taking a source-sink simulation with recycling events and setting up a new simulation in the opposite direction (sink -> source).

Slimygen:
I am working on updating Darian's prototype to be compatible with modern WESTPA.

PR #587, opened 2026-05-01
Adding w_reverse to the develop branch

Slimygen:
## Issue Number

#543

## Describe the changes made

Adding a tool called w_reverse to easily create basis states for a WE run from an already completed run of the reverse process.

- [ ] Make sure w_reverse works no matter how the restart files are saved
- [ ] Make tests
- [ ] Update documentation

## Major files changed

- [ ] setup.py
- [ ] src/westpa/cli/tools/w_reverse.py

## Status

- [ ] My code follows the code style of this project.
- [ ] I have read the AI POLICY document.
- [ ] This PR conforms to WESTPA's AI policy.
- [ ] I have declared all usage of AI in the PR description.


src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+from tqdm.auto import tqdm
+import os
+import shutil
+import argparse
+from io import BytesIO
+import tarfile
+from westpa.core.h5io import WESTIterationFile
+from westpa.core.h5io import safe_extract
+from westpa.core._rc import WESTRC
+
```
jeremyleung521, line 13:
```suggestion
from westpa.tools import (
    WESTTool,
    WESTDataReader,
    IterRangeSelection,
)
```

Use these built-in parts instead to better integrate with other WESTPA tools. This will also help if it'll eventually be upgraded to run in parallel or sth.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+log = logging.getLogger('w_reverse')
+
+
+class W_Reverse:
```
jeremyleung521:
```suggestion
class W_Reverse(WESTTool):
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            Default './bstates_reverse'.'
+        output_bstates_file : str
+            Name of the output bstates file, default 'bstates.txt'.
+        use_weights : bool
+            By default, include the recycled event weight when making the bstates.txt file.
+            temp_dir : str
+        Name of the temporary directory that will be created
+        """
+        # Parse config file
+        westrc = WESTRC()
```
jeremyleung521:
```suggestion
        westrc = rc or westpa.rc
```

respect passed rc if exists, else make one.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            Output directory for the bstates and output_bstates_file.
+            Default './bstates_reverse'.'
+        output_bstates_file : str
+            Name of the output bstates file, default 'bstates.txt'.
+        use_weights : bool
+            By default, include the recycled event weight when making the bstates.txt file.
+            temp_dir : str
+        Name of the temporary directory that will be created
+        """
+        # Parse config file
```
jeremyleung521: set config_required so the subclassed `WESTTool` will read the config file.
```suggestion
        self.config_required = True
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+                    else:
+                        bstates_f.write(f"{idx} 1 {rst_dest_name}\n")
+                    n_bstates += 1
+                else:
+                    break
+            # Remove temporary directory
+            shutil.rmtree(self.temp_dir)
+
+
+def parse_arguments():
```
jeremyleung521:
```suggestion
Like other tools, I would put this under `W_Reverse.add_args()`
```
https://github.com/westpa/westpa/blob/westpa2/src/westpa/cli/tools/w_ipa.py#L81

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            temp_dir : str
+        Name of the temporary directory that will be created
+        """
+        # Parse config file
+        westrc = WESTRC()
+        westrc.read_config(config_file)
+        self.config_file = config_file
+        config = westrc.config
+        # Read the west.h5 file
+        self.h5 = h5py.File(h5, mode="r")
```
jeremyleung521: This seems like a lot at the `__init__` phase. I would actually break it up to a separate `process_args()` block, like other tools

https://github.com/westpa/westpa/blob/3a213bbbdec95a08e51ca9457513f69bd1908ef0/src/westpa/cli/tools/w_ipa.py#L113

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+
+    TODO:
+        * option to use w_assign/assign.h5 output for successful bstate selection
+        * adapt for WE simulation traj_segs that used the hdf5 framework
+        * integrate into WESTPA, pull instance attributes and args from west.cfg
+        * change printing to west logging
+    """
+
+    def __init__(
+        self,
```
jeremyleung521:
```suggestion
        self,
        rc=None,
```

src/westpa/cli/tools/w_reverse.py
jeremyleung521: I would defer this to the `WESTDataReader` 

https://github.com/westpa/westpa/blob/westpa2/src/westpa/cli/tools/w_assign.py#L283

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+        self.output_bstates_file = str(output_bstates_file)
+        self.use_weights = use_weights
+
+    def w_succ(self):
+        """
+        Find and return all successfully recycled (iter, seg) pairs.
+
+        Returns
+        -------
+        succ : list of tuples (iter,wlk,weight)
```
jeremyleung521: be more precise about `wlk`. I think we prefer more verbose variable names, so use `iter_id`, `seg_id` etc...

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+        succ = []
+
+        for iter in tqdm(range(self.first_iter, self.last_iter + 1), desc="Running w_succ"):
+            # if the new_weights group exists in the h5 file
+            if f"iterations/iter_{iter:08d}/new_weights" in self.h5:
+                prev_segs = self.h5[f"iterations/iter_{iter:08d}/new_weights/index"]["prev_seg_id"]
+                recycled_weights = self.h5[f"iterations/iter_{iter:08d}/new_weights/index"]["weight"]
+                # append the previous iter and previous seg id recycled and the weight
+                for i in range(len(prev_segs)):
+                    succ.append((iter - 1, prev_segs[i], recycled_weights[i]))
```
jeremyleung521: I feel like we could do this more straightforwardly by looking at `seg_index` status code directly instead of from new_wieghts and then figuring backwards (the index there is exactly the seg_id). Could turn the `for...append` to a list comprehension too for a small speed up, then vstack it to a full numpy array.

w_succ code:
https://github.com/westpa/westpa/blob/3a213bbbdec95a08e51ca9457513f69bd1908ef0/src/westpa/cli/core/w_succ.py#L45

westpa.analysis code (based on w_succ):
https://github.com/westpa/westpa/blob/3a213bbbdec95a08e51ca9457513f69bd1908ef0/src/westpa/analysis/core.py#L281

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            print(f"Directory '{directory}' created.")
+        else:
+            print(f"Directory '{directory}' already exists.")
+
+    def w_reverse(self):
+        """
+        Main public method for running w_reverse.
+        """
+        # Start a new instance of WESTRC to parse the config file
+        westrc = WESTRC()
```
jeremyleung521: You're building yet another (local) WESTRC() here. store the config as `self.config`

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+                        except ValueError as e:
+                            log.warning(f'could not write HDF5 Framework restart data for iteration {it} walker {wlk}: {e}')
+                            if it == 1:
+                                log.warning(
+                                    f'In iteration 1. Assuming this is a start state and proceeding to skip reading restart from per-iteration HDF5 file for iteration {it} walker {wlk}'
+                                )
+                        except Exception as e:
+                            log.warning(f'could not write HDF5 Framework restart data for iteration {it} walker {wlk}: {e}')
+                        # Look at all files in the temp directory
+                        temp_dir_contents = os.listdir(self.temp_dir)
```
jeremyleung521: watch out for hidden files with a "." in front of it.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+                        extention_not_found = True
```
jeremyleung521:
```suggestion
                        extension_not_found = True
```
typo. Change the rest...

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            self.h5_framework = True
+        else:
+            traj_seg_path_list = data_refs_dic['segment'].split('/')[1:]
+        self.traj_segs_path = '/'.join(traj_seg_path_list)
+        self.max_n_bstates = int(max_n_bstates)
+        self.rst_file = str(rst_file)
+        # Get the restart file extention being used
+        self.rst_extension = self.rst_file.split('.')[-1]
+        self.output_bstates_dir = str(output_bstates_dir)
+        self.temp_dir = str(temp_dir)
```
jeremyleung521: Not sure why we're not making a `tempfile.TemporaryDirectory` instead. It's for untarring the hdf5 file tars, right? That way you don't have to worry about recursively overwriting files on top of each other and cleanup (use context managers).

https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+                    # bstates.txt row format: bstate_n | weight | bstate_filename
+                    if self.use_weights:
+                        bstates_f.write(f"{idx} {weight:.3e} {rst_dest_name}\n")
+                    else:
+                        bstates_f.write(f"{idx} 1 {rst_dest_name}\n")
+                    n_bstates += 1
+                else:
+                    break
+            # Remove temporary directory
+            shutil.rmtree(self.temp_dir)
```
jeremyleung521: If this loop fails halfway, this will not run. Use a `TemporaryDirectory` context manager. or `try`/`finally` block.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+                    rst_dest_name = f"{it:06d}_{wlk:06d}.{self.rst_extension}"
+                    # check if using HDF5 framework
+                    if self.h5_framework:
+                        # Find how the .h5 files for each iteration are named
+                        traj_seg_file_name = data_refs_dic["iteration"].split('/')[-1]
+                        # Make a path to this iterations .h5 file
+                        traj_seg = f'{self.traj_segs_path}/{traj_seg_file_name}'
+                        # Extracct the restart data from the .h5 file
+                        h5file = WESTIterationFile(traj_seg.format(n_iter=it))
+                        restart_data = h5file.read_data('/restart/%d_%d' % (it, wlk), 'data')
```
jeremyleung521: Wouldn't it be better to directly use the `restart_writer` function here? They were separated out for that reason.

https://github.com/westpa/westpa/blob/3a213bbbdec95a08e51ca9457513f69bd1908ef0/src/westpa/core/propagators/loaders.py#L118

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,305 @@
...
+            # different totals if the max is less than total succ_pairs to loop
+            if self.max_n_bstates < len(succ_pairs):
+                total_pairs = self.max_n_bstates
+            else:
+                total_pairs = len(succ_pairs)
+
+            # then for each pair
+            for idx, (it, wlk, weight) in tqdm(enumerate(succ_pairs), total=total_pairs, desc="New bstates"):
+                # Make sure you are not over the maximum bstates
+                if n_bstates < self.max_n_bstates:
```
jeremyleung521: This is a little weird... You're prioritizing the earlier recycles over the later ones. It would actually be easier to actually use 

```
rng = np.random.default_rng()
idx = rng.choice(
    self.len(succ_pairs), 
    p=succ_pairs[:, 2] / np.sum(succ_pairs[:, 2], dtype=float),
    replace=False
)
```
to randomly select based on normalized trajectory weights, then use those selected indices.


https://numpy.org/doc/stable/reference/random/generated/numpy.random.Generator.choice.html#numpy.random.Generator.choice

jeremyleung521 (changes requested):
Some stylistic/code changes to keep in line with current WESTPA code + one typo + one major change in selection criteria if number of succ trajs >> max_n_bstates

Also some tests are needed. :)

jeremyleung521:
Another feature request (I guess) is that the "supply the restart file name" portion could theoretically be automated with this function if not supplied.

https://github.com/westpa/westpa/blob/westpa2/src/westpa/core/trajectory.py#L408

PR #594, opened 2026-06-01
GSoC 2026: Interface for post-simulation analysis (“crawling”) of WESTPA simulations

kunjsinha:
## Describe the changes made
This PR contains the files for the GSoC 2026 crawler project (https://summerofcode.withgoogle.com/programs/2026/projects/VJMrp2UW).

A rough timeline can be found at https://kunjsinha.github.io/2026/05/13/GSoC-intro.html

## Major files changed

- [ ]

## Status

- [ ] My code follows the code style of this project.
- [ ] I have read the AI POLICY document.
- [ ] This PR conforms to WESTPA's AI policy.
- [ ] I have declared all usage of AI in the PR description.


src/westpa/core/mdacrawl.py
```diff
@@ -0,0 +1,93 @@
+'''Parser and Reader to expose WESTPA simulation data as MDAnalysis Universe'''
+
+import json
+
+import h5py
+import numpy as np
+
+from MDAnalysis.topology.base import TopologyReaderBase
```
jeremyleung521: I know it's not PEP8-compliant but since MDA is not a hard dependency of WESTPA, I'd put most of the imports within the WESTPAParser Class. Also would save some import time right off the bat.

TopologyReaderBase specifically though... you might have to do this:
https://stackoverflow.com/a/42297227

jeremyleung521:
Feel free to pull `develop`. There were a few things that broke in the CI and the latest commit(s) should fix them.

src/westpa/core/mdanalysis_core.py
```diff
@@ -0,0 +1,203 @@
...
+import warnings
+
+import h5py
+import numpy as np
+
+try:
+    from MDAnalysis.topology.base import TopologyReaderBase
+
+    MDATopologyBase = TopologyReaderBase
+except ImportError:
```
jeremyleung521, line 13: I believe this would throw a `ModuleNotFoundError`, not an `ImportError`

src/westpa/core/mdanalysis_core.py
```diff
@@ -0,0 +1,203 @@
...
+    MDATopologyBase = object
+
+
+class WESTPAParser(MDATopologyBase):
+    format = 'WESTPA'
```
jeremyleung521, line 18: might be worth adding in a version number as well.

src/westpa/core/mdanalysis_core.py
```diff
@@ -0,0 +1,203 @@
...
+            atom_resindex=np.array(atom_resindex, dtype=np.int32),
+            residue_segindex=np.array(residue_segindex, dtype=np.int32),
+        )
+
+
+try:
+    from MDAnalysis.coordinates.base import ReaderBase
+
+    MDAReaderBase = ReaderBase
+except ImportError:
```
jeremyleung521, line 120: `ModuleNotFoundError`, see above.

src/westpa/core/mdanalysis_core.py
```diff
@@ -0,0 +1,203 @@
...
+
+class WESTPAReader(MDAReaderBase):
+    format = 'WESTPA'
+
+    def __init__(self, filename, n_atoms=None, **kwargs):
+        self._n_atoms = n_atoms
+        super().__init__(filename, **kwargs)
+        self.filename = filename
+
+        self._h5 = h5py.File(filename, 'r')
```
jeremyleung521, line 133: Not sure if we should keep this file hanging.

src/westpa/core/mdacrawl.py
```diff
@@ -0,0 +1,203 @@
...
+
+        self._h5 = h5py.File(filename, 'r')
+        # Will be used for caching later
+        self._current_h5 = None
+        self._current_path = None
+
+        self.iter_prec = self._h5.attrs['west_iter_prec']
+
+        self.frame_index = []
+        n_iters = self._h5['summary'].shape[0]
```
jeremyleung521: I usually do a `len([h5['summary']['walltime', : ] != 0])` mask because there might be a prepped iteration with no data. Not sure if that's better to do here.

jeremyleung521 (commented):
Just some comments. Will add more later.

tests/test_mdacrawl.py
jeremyleung521, line 1: since mdacrawl is a cli tool, it should probably sit within `test_tools`.

tests/test_mdacrawl.py
```diff
@@ -0,0 +1,262 @@
...
+        if len(oxygen) == 1:
+            bonded_atoms = oxygen[0].bonded_atoms
+            assert np.all(np.isin(bonded_atoms.names, ['H1', 'H2', 'HW1', 'HW2'])), "Oxygen bonded to non-hydrogen atom"
+
+
+class Test_WESTPAReader:
+    """Class to test the WESTPAReader dynamic trajectory reading module"""
+
+    @pytest.fixture(scope="class")
+    def mda_universe(self):
```
jeremyleung521: This deprecation error showed up. 
https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method
```
tests/test_mdacrawl.py::Test_WESTPAParser::test_parser_atom_count
tests/test_mdacrawl.py::Test_WESTPAReader::test_reader_frame_count
  /home/jml230/apps/miniforge3/envs/westpa-mdacrawl/lib/python3.13/site-packages/_pytest/fixtures.py:1313: PytestRemovedIn10Warning: Class-scoped fixture defined as instance method is deprecated.
  Instance attributes set in this fixture will NOT be visible to test methods,
  as each test gets a new instance while the fixture runs only once per class.
  Use @classmethod decorator and set attributes on cls instead.
  See https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method
    fixturefunc = resolve_fixture_function(fixturedef, request)
```


```suggestion
    @pytest.fixture(scope="class")
    @classmethod
    def mda_universe(cls):    
```

tests/test_mdacrawl.py
```diff
@@ -0,0 +1,262 @@
...
+
+here = os.path.dirname(os.path.abspath(__file__))
+hdf5_file = os.path.join(here, 'refs', 'west_mdacrawl.h5')
+
+
+class Test_WESTPAParser:
+    """Class to test the WESTPAParser topology reading module"""
+
+    @pytest.fixture(scope="class")
+    def mda_universe(self):
```
jeremyleung521: This deprecation error showed up. 
https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method
```
tests/test_mdacrawl.py::Test_WESTPAParser::test_parser_atom_count
tests/test_mdacrawl.py::Test_WESTPAReader::test_reader_frame_count
  /home/jml230/apps/miniforge3/envs/westpa-mdacrawl/lib/python3.13/site-packages/_pytest/fixtures.py:1313: PytestRemovedIn10Warning: Class-scoped fixture defined as instance method is deprecated.
  Instance attributes set in this fixture will NOT be visible to test methods,
  as each test gets a new instance while the fixture runs only once per class.
  Use @classmethod decorator and set attributes on cls instead.
  See https://docs.pytest.org/en/stable/deprecations.html#class-scoped-fixture-as-instance-method
    fixturefunc = resolve_fixture_function(fixturedef, request)
```


```suggestion
    @pytest.fixture(scope="class")
    @classmethod
    def mda_universe(cls):    
```

src/westpa/core/mdanalysis_core.py
jeremyleung521, line 1: Would it be better to rename this as `mdanalysis_core.py` or `mda_core.py` instead, considering these parts may potentially be reused outside of the scope of mdacrawl?

jeremyleung521 (commented):
Just some quick comments about organization. Will add more as I dig deeper.

src/westpa/core/mdanalysis_core.py
kunjsinha: Yes, that also works! Right now the name just refers to the tool's use case but it can be generalized incase it has to be used outside its scope or expand on in the future. `mdanalysis_core.py` seems good.

tests/test_mdacrawl.py
```diff
@@ -277,6 +278,8 @@ def test_saved_auxdata_compatability(self, mda_universe, tmp_path):
 
         # Slice as [:, 2:3] instead of [:, 2] so that save_to_west_h5() is forced to build (Segment, Frames, 1), satisfying w_assign
         flat_results = R.results.rmsd[:, 2:3]
```
jeremyleung521: `R.results.rmsd[:, 2, None]` is probably the more "correct" way to do this. "None" is a way to call `np.newaxis()`

tests/test_mdacrawl.py
```diff
@@ -277,6 +278,8 @@ def test_saved_auxdata_compatability(self, mda_universe, tmp_path):
...
+        flat_results = np.nan_to_num(flat_results, nan=0.0)
+        flat_results = np.clip(flat_results, a_min=0.0, a_max=None)
```
jeremyleung521: This will probably lead to false negatives, where failed calculations (NaN) are replaced and ignored, hiding errors instead of exposing them for fixing.

PR #605, opened 2026-07-28
Adding a w_reverse tool

Slimygen:
## Issue Number

#543

## Describe the changes made

Adding a cli tool to start new simulations from the successful trajectories of previous simulations. This will make it easier to simulate forward and reverse processes one after the other.

- [x] Make w_reverse tool
- [x] Make tests for w_reverse tool
- [x] Update Documentation

## Major files changed

- [x] /src/westpa/cli/tools/w_reverse.py
- [x] /tests/test_tools/test_w_reverse.py
- [x] Added numerous reference files needed for testing

## Status

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] I have read the AI POLICY document.
- [x] This PR conforms to WESTPA's AI policy.
- [x] I have declared all usage of AI in the PR description.


src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+        with open(os.path.join(self.output_bstates_dir, self.output_bstates_file), "w") as bstates_f:
+
+            # Number of reverse bstates created
+            # different totals if the max is less than total succ_pairs to loop
+            if self.max_n_bstates < len(succ_pairs):
+                total_pairs = self.max_n_bstates
+            else:
+                total_pairs = len(succ_pairs)
+            # then for each pair
+            rng = random.default_rng(12345)
```
jeremyleung521: We shouldn't be hardcoding the rng with a seed. In fact, making it an user-facing argument would probably be the best choice for users to reproduce answers. `log.info()` the seed out into the stdout too.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+        os.makedirs(self.output_bstates_dir, exist_ok=True)
+        # create bstates.txt file
+        with open(os.path.join(self.output_bstates_dir, self.output_bstates_file), "w") as bstates_f:
+
+            # Number of reverse bstates created
+            # different totals if the max is less than total succ_pairs to loop
+            if self.max_n_bstates < len(succ_pairs):
+                total_pairs = self.max_n_bstates
+            else:
+                total_pairs = len(succ_pairs)
```
jeremyleung521:
```suggestion
            total_pairs = min(self.max_n_bstates, len(succ_pairs))
```
Probably the cleanest way for this.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+                    # os.listdir(seg_path)
+                    rst_file_path = os.path.join(seg_path, self.rst_file)
+                    shutil.copyfile(rst_file_path, os.path.join(self.output_bstates_dir, rst_dest_name))
+                # fill out the bstates.txt file with name and weight
+                # but only use weights if requested, otherwise use equal weights
+                # bstates.txt row format: bstate_n | weight | bstate_filename
+                if self.use_weights:
+                    bstates_f.write(f"{idx} {weight:.3e} {rst_dest_name}\n")
+                else:
+                    bstates_f.write(f"{idx} 1 {rst_dest_name}\n")
```
jeremyleung521:
```suggestion
                    bstates_f.write(f"{idx} {1/total_pairs:.3e)} {rst_dest_name}\n")
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+                walker = int(succ_pairs[index][1])
+                weight = float(succ_pairs[index][2])
+                # check if using HDF5 framework
+                if self.h5_framework:
+                    with tempfile.TemporaryDirectory() as tmpdirname:
+                        # Extracct the restart data from the .h5 file
+                        h5file = WESTIterationFile(self.traj_seg.format(n_iter=iteration))
+                        restart_data = h5file.read_data('/restart/%d_%d' % (iteration, walker), 'data')
+                        segment = Segment(n_iter=iteration, seg_id=walker, weight=weight)
+                        segment.data['iterh5/restart'] = restart_data
```
jeremyleung521:
```suggestion
                        # Extract the restart data from the .h5 file
                        segment = Segment(n_iter=iteration, seg_id=walker, weight=weight)
                        h5file = WESTIterationFile(self.traj_seg.format(n_iter=iteration))
                        h5file.read_restart(segment)
```

This will `ValueError` out nicely if restart does not exist. Also typo.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+                        restart_writer(tmpdirname, segment)
+                        # Look at all files in the temp directory
+                        if self.rst_file:
+                            rst_dest_name = f"{iteration:06d}_{walker:06d}.{self.rst_extension}"
+                            temp_dir_contents = os.listdir(tmpdirname)
+                            file_not_found = True
+                            for temp_file in temp_dir_contents:
+                                # Move and rename only the restart file with the user specified extension to the bstate directory
+                                if temp_file.lower() == self.rst_file:
```
jeremyleung521:
```suggestion
                                if temp_file.lower().endswith(self.rst_extension):
```


1. Instead of having to make the user to forcefully type out the file name, pick the file based on extension would make the logic much easier. 
2. In fact, you could probably remove the `file_not_found` flag (and the next two `if` blocks) and clean up the whole search loop, which should work with both hdf5 framework + non-case if you're in the correct folder (traj_segs or untarred folder). No need for different "cases":

e.g.
``` 
search_folder = tmpdirname if self.h5_framework else 
                        makepath(traj_segs_path, {'data_dir': tmpdirname, 'segment': segment}
_ , temp_file = find_top_traj_file(search_folder, [], [self.rst_extension])
shutil.move(temp_file,  os.path.join(self.output_bstates_dir, rst_dest_name)
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+                        .format(n_iter=iteration, seg_id=walker)
+                    )
+                    # os.listdir(seg_path)
+                    rst_file_path = os.path.join(seg_path, self.rst_file)
+                    shutil.copyfile(rst_file_path, os.path.join(self.output_bstates_dir, rst_dest_name))
+                # fill out the bstates.txt file with name and weight
+                # but only use weights if requested, otherwise use equal weights
+                # bstates.txt row format: bstate_n | weight | bstate_filename
+                if self.use_weights:
+                    bstates_f.write(f"{idx} {weight:.3e} {rst_dest_name}\n")
```
jeremyleung521: I'm a little back and forth between directly using `weight/sum(weights)` vs. just using the traj weight and letting w_init normalize it (what you're doing here).

Personally, holding the output file open the whole time through tons of copying/query is a little risky and i'd actually loop through `indices` a second time just to write the bstate file, to be safe.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+                        # Extracct the restart data from the .h5 file
+                        h5file = WESTIterationFile(self.traj_seg.format(n_iter=iteration))
+                        restart_data = h5file.read_data('/restart/%d_%d' % (iteration, walker), 'data')
+                        segment = Segment(n_iter=iteration, seg_id=walker, weight=weight)
+                        segment.data['iterh5/restart'] = restart_data
+                        restart_writer(tmpdirname, segment)
+                        # Look at all files in the temp directory
+                        if self.rst_file:
+                            rst_dest_name = f"{iteration:06d}_{walker:06d}.{self.rst_extension}"
+                            temp_dir_contents = os.listdir(tmpdirname)
```
jeremyleung521:
```suggestion
                            temp_dir_contents = [f_name for f_name in os.listdir(tmpdirname) if not f_name.startswith('.')]
```

You could get tmp/swap hidden files if you don't do this.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+        for iteration_index, iteration in tqdm(
+            enumerate(self.h5['iterations'].keys()), total=len(self.h5['iterations'].keys()), desc="w_succ"
+        ):
+            endpoint_type = self.h5['iterations'][iteration]['seg_index']['endpoint_type']
+            indices = flatnonzero(endpoint_type == Segment.SEG_ENDPOINT_RECYCLED)
+            temp_array = [
+                [iteration_index + 1, index, self.h5['iterations'][iteration]['seg_index']['weight'][index]] for index in indices
+            ]
+            succ += temp_array
+        return array(succ)
```
jeremyleung521:
```suggestion
        return np.asarray(succ)
```
Want to limit the amount of copying as possible.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+            # Number of reverse bstates created
+            # different totals if the max is less than total succ_pairs to loop
+            if self.max_n_bstates < len(succ_pairs):
+                total_pairs = self.max_n_bstates
+            else:
+                total_pairs = len(succ_pairs)
+            # then for each pair
+            rng = random.default_rng(12345)
+            indices = rng.choice(
+                len(succ_pairs), size=total_pairs, p=succ_pairs[:, 2] / sum(succ_pairs[:, 2], dtype=float), replace=False
```
jeremyleung521: Is there a reason why you're running `rng.choice` on `len(succ_pairs)` and not `succ_pairs` directly? If `succ_pairs` is large, querying it once in `rng.choice` could potentially could be faster than having to query it in L229-L231 multiple times.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+            if self.data_refs_dic['iteration'][0] == '/':
+                starts_with_slash = True
+            self.h5_framework = True
+        else:
+            if self.data_refs_dic['segment'][0] == '$':
+                traj_seg_path_list = self.data_refs_dic['segment'].split('/')[1:]
+            else:
+                traj_seg_path_list = self.data_refs_dic['segment'].split('/')
+            if self.data_refs_dic['segment'][0] == '/':
+                starts_with_slash = True
```
jeremyleung521: This whole thing is just to remove $WEST_SIM_ROOT? Seems like a maintenance nightmare... and a ton of assumptions (it's at the start, it's the only environment variable). You can first check if $WEST_SIM_ROOT is an environment variable (or if provided as an argument), just use [makepath](https://github.com/westpa/westpa/blob/westpa2/src/westpa/core/data_manager.py#L77
) to replace, else implicitly replace with a different placeholder (`{data_dir}`) to be replaced later on.

E.g., 
```suggestion
            self.h5_framework = True if 'iteration in self.data_refs_dic else False
            dict_key = 'iteration' if self.h5_framework  else 'segment'

            traj_seg_path_list = makepath(self.data_refs_dic[dict_key])   # Check kwargs to see what you want
            
            # Proves that WEST_SIM_ROOT is not on os.environ, not replaced
            if '$WEST_SIM_ROOT' in traj_seg_path_list:
                traj_seg_path_list = traj_seg_path_list.replace('$WEST_SIM_ROOT', '{data_dir}')
```

and then later on when you need it:

```
makepath(traj_seg_path_list, {'segment': segment, 'data_dir': 'WHATEVER PATH'})
```

jeremyleung521 (changes requested):
Some changes for maintainability. Sorry it's a lot. I haven't touched the tests just yet.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,300 @@
...
+from tqdm.auto import tqdm
+import os
+import shutil
+import tempfile
+from westpa.core.h5io import WESTIterationFile, WESTPAH5File
+from westpa.core.propagators.loaders import restart_writer
+from westpa.core.segment import Segment
+from westpa.core._rc import WESTRC
+from westpa.tools import WESTTool, WESTDataReader
+from numpy import flatnonzero, random, array, sum
```
jeremyleung521: I'm against importing these out of numpy like this and then using them... Having them as `np.___` underneath actually lets the reader understand we're dealing with numpy arrays and functions. Not to mention, you're overwriting `sum` (python default function) with numpy's version.
```suggestion
import numpy as np
```

Then in the code:
```
np.sum(variable)
```

src/westpa/cli/tools/w_reverse.py
Slimygen: In the case that the user is not using the hdf5 frame work then the parent restart file and the segment restart files are stored in the same directory. Without specifying a file name how would w_reverse tell which to use?

jeremyleung521: > In the case that the user is not using the hdf5 frame work then the parent restart file and the segment restart files are stored in the same directory. Without specifying a file name how would w_reverse tell which to use?

Ah. yea that makes sense. Forgot about that.

Then the priority should be:
1. Search by name
2. Search by extension
3. If multiple files hit: search for latest creation date or something? 

```
import os
from os.path import getctime

search_folder = tmpdirname if self.h5_framework else 
                        makepath(traj_segs_path, {'data_dir': tmpdirname, 'segment': segment}

files = [file for file in os.listdir(search_folder) if not file.startswith('.')]

if self.rst_file in files:
    shutil.move(self.rst_file,  os.path.join(self.output_bstates_dir, rst_dest_name))
else:
    possible_hits = [file for file in files if file.endswith(self.rst_extension)]
    if len(possible_hits) > 1:
        possible_hits = sorted(possible_hits, key=lambda file: getctime(file))
        log.warning(f'Found {possible_hits[-1]} as restart file for iteration ... based on file creation times. if this is incorrect, provide a file name...')

    shutil.move(possible_hits[-1],  os.path.join(self.output_bstates_dir, rst_dest_name))

```

tests/test_tools/test_w_reverse.py
```diff
@@ -0,0 +1,63 @@
...
+    test_name = 'W_Reverse'
+
+    def test_run_w_reverse_no_hdf5(self):
+        '''Testing if w_reverse runs as expected and the h5 files looks good.'''
+
+        ref_dir = os.path.join(os.path.dirname(__file__), '../refs')
+        shutil.copy2(os.path.join(ref_dir, 'west_reverse_no_hdf5.cfg'), './west.cfg')
+        shutil.copy2(os.path.join(ref_dir, 'west_reverse_no_hdf5.h5'), './west.h5')
+        shutil.copytree(os.path.join(ref_dir, 'traj_segs_reverse_no_hdf5'), './traj_segs')
+        os.system('w_reverse --rst-file seg.xml')
```
jeremyleung521: Ah yea. Don't do this (`os.system`). Call the python functions (`entry_point` or what not) so pytest actually knows you're running through those parts of the code. Makes sense why code coverage is not catching anything.

Something like this:
https://github.com/westpa/westpa/blob/westpa2/tests/test_tools/test_w_multi_west.py#L17-L33

tests/test_tools/test_w_reverse.py
```diff
@@ -0,0 +1,63 @@
...
+        assert os.path.isfile('./bstates_reverse/bstates.txt'), "The bstates.txt file was not generated."
+        assert os.path.isfile('./bstates_reverse/000001_000000.xml'), "The 000001_000000.xml file was not generated."
```
jeremyleung521, line 38: and maybe that the file is not empty?

tests/test_tools/test_w_reverse.py
```diff
@@ -0,0 +1,63 @@
+import os
+import shutil
+import unittest
+from filecmp import cmp
+
+
+class Test_W_Reverse(unittest.TestCase):
+    test_name = 'W_Reverse'
+
+    def test_run_w_reverse_no_hdf5(self):
```
jeremyleung521, line 12: Work/Use tmpdir/tmp_path fixtures. Will save us from a lot of the potential conflicts with other tests. In fact, if any of these tests require same setups, make a fixture in conftest!

https://docs.pytest.org/en/4.6.x/tmpdir.html#the-tmpdir-fixture
https://github.com/westpa/westpa/blob/westpa2/tests/conftest.py#L37

tests/refs/traj_segs_reverse_no_hdf5/000001/000000/seg.xml
jeremyleung521, line 1: We really don't need this many atoms for a test file. A 5-10 atom file should suffice. In fact, I don't even think we need any content in these test files, (just dummy files with the right extensions) since we are shutil.copyfile() wholesale.

src/westpa/cli/tools/w_reverse.py
jeremyleung521: I would also do `shutil.copyfile()` instead. incomplete `moves`  are messy.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -90,11 +90,15 @@ def add_args(self, parser):
...
             "-nw",
+            type=bool,
+            default=True,
             dest="use_weights",
-            action="store_false",
-            help="Don't include the recycled event weight when making the bstates.txt file",
+            help="Include the recycled event weight when making the bstates.txt file",
+        )
+        rgroup.add_argument(
+            '--seed', type=int, default=12345, dest='seed', help='Seed for randomly choosing which states to include'
```
jeremyleung521: default, imo, should be None, so it's stochastic every time, unless prompted otherwise.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+            type=str,
+            default="bstates.txt",
+            help="Name of the output bstates file",
+        )
+        # TODO: may need to be adjusted to store False when included
+        rgroup.add_argument(
+            "--use-weights",
+            "-nw",
+            type=bool,
+            default=True,
```
jeremyleung521:
```suggestion
            action='store_false',
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        )
+        rgroup.add_argument(
+            "--output-bstates-file",
+            "-obf",
+            dest="output_bstates_file",
+            type=str,
+            default="bstates.txt",
+            help="Name of the output bstates file",
+        )
+        # TODO: may need to be adjusted to store False when included
```
jeremyleung521:
```suggestion
```
Just do it.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+
+    def process_args(self, args):
+        self.data_reader.process_args(args)
+        self.config_required = True
+        self.config_file = args.config_file
+        self.westrc.read_config(self.config_file)
+        self.config = self.westrc.config
+        # Read the west.h5 file
+        self.h5 = WESTPAH5File(args.we_h5filename, 'r')
+        self.first_iter = int(args.first_iter)
```
jeremyleung521:
```suggestion
        self.first_iter = args.first_iter
```
Argparse already does the type conversion for you. Letting argparse do it will also prevent cases like `int(None)`.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        # default to last
+        if args.last_iter is not None:
+            self.last_iter = int(args.last_iter)
```
jeremyleung521:
```suggestion
            self.last_iter = args.last_iter
```

See comment above

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        elif args.last_iter is None:
+            self.last_iter = self.h5.attrs['west_current_iteration'] - 1
+        # Look at the data_refs from the config file
+        self.data_refs_dic = self.config['west']['data']['data_refs']
```
jeremyleung521:
```suggestion
        self.data_refs_dic = self.config.get(['west', 'data', 'data_refs'], {})
```

Grab in one swoop, to prevent caching of the intermediate dictionaries. Also will fail more gracefully by returning an empty dictionary.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        self.westrc.read_config(self.config_file)
+        self.config = self.westrc.config
+        # Read the west.h5 file
+        self.h5 = WESTPAH5File(args.we_h5filename, 'r')
+        self.first_iter = int(args.first_iter)
+        # default to last
+        if args.last_iter is not None:
+            self.last_iter = int(args.last_iter)
+        elif args.last_iter is None:
+            self.last_iter = self.h5.attrs['west_current_iteration'] - 1
```
jeremyleung521:
```suggestion
        self.last_iter = args.last_iter or self.h5.attrs['west_current_iteration'] - 1
```
Note that `west_current_iteration` _could be_ a completed iteration if you ran `w_truncate`

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        self.data_refs_dic = self.config['west']['data']['data_refs']
+        # Default to not using HDF5 framework
+        self.h5_framework = True if 'iteration' in self.data_refs_dic else False
+        dict_key = 'iteration' if self.h5_framework else 'segment'
+        self.traj_seg_path = os.path.expandvars(
+            self.data_refs_dic[dict_key].replace('segment.n_iter', 'n_iter').replace('segment.seg_id', 'seg_id')
+        )
+        # Proves that WEST_SIM_ROOT is not on os.environ, not replaced
+        self.traj_seg_path = self.traj_seg_path.replace('$WEST_SIM_ROOT', '.')
+        self.max_n_bstates = int(args.max_n_bstates)
```
jeremyleung521:
```suggestion
        self.max_n_bstates = args.max_n_bstates
```

see comment above

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        if args.rst_file:
+            self.rst_file = str(args.rst_file).lower()
+            # Get the restart file extension being used
+            self.rst_extension = self.rst_file.split('.')[-1]
+        else:
+            self.rst_file = None
+            self.rst_extension = None
```
jeremyleung521:
```suggestion
        self.rst_file = args.rst_file.lower() if args.rst_file else None
        self.rst_extension = self.rst_file.rsplit('.', maxsplit=1)[-1] if self.rst_file else None
```

Prefer more concise declarations.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        self.traj_exc_exts = []
+        self.traj_or_top_exts = []
+        for i in self.traj_exts:
+            if i in self.top_exts:
+                self.traj_or_top_exts.append(i)
+            else:
+                self.traj_exc_exts.append(i)
+        self.output_bstates_dir = str(args.output_bstates_dir)
```
jeremyleung521:
```suggestion
        self.output_bstates_dir = args.output_bstates_dir
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        self.output_bstates_file = str(args.output_bstates_file)
```
jeremyleung521:
```suggestion
        self.output_bstates_file = args.output_bstates_file
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        for iteration_index, iteration in tqdm(
+            enumerate(self.h5['iterations'].keys()), total=len(self.h5['iterations'].keys()), desc="w_succ"
+        ):
+            endpoint_type = self.h5['iterations'][iteration]['seg_index']['endpoint_type']
+            indices = np.flatnonzero(endpoint_type == Segment.SEG_ENDPOINT_RECYCLED)
+            temp_array = [
+                [
+                    iteration_index if self.h5_framework else iteration_index + 1,
+                    index,
+                    self.h5['iterations'][iteration]['seg_index']['weight'][index],
```
jeremyleung521:
```suggestion
                    self.h5[f'iterations/{iteration}/'seg_index']['weight', index],
```
Will prevent creating intermediate copies

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+
+        Returns
+        -------
+        succ : array of shape (n, 3) with [iteration, walker, weight] for each succ[i]
+        """
+        succ = []
+        for iteration_index, iteration in tqdm(
+            enumerate(self.h5['iterations'].keys()), total=len(self.h5['iterations'].keys()), desc="w_succ"
+        ):
+            endpoint_type = self.h5['iterations'][iteration]['seg_index']['endpoint_type']
```
jeremyleung521:
```suggestion
            endpoint_type = self.h5[f'iterations/{iteration}/seg_index']['endpoint_type', :]
```

See below

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        Returns
+        -------
+        rst_dest_name: string describing the name of the copied trajectory in the output_bstates_dir
+        '''
+        files = [file for file in os.listdir(search_folder) if not file.startswith('.')]
+        if self.rst_file in files:
+            rst_dest_name = f"{iteration:06d}_{walker:06d}.{self.rst_extension}"
+            shutil.copyfile(os.path.join(search_folder, self.rst_file), os.path.join(self.output_bstates_dir, rst_dest_name))
+            return rst_dest_name
+        if self.rst_extension:
```
jeremyleung521:
```suggestion
        elif self.rst_extension:
```

More a stylistic choice than anything. The `elif` better implies right off the bat that other blocks won't be running, which is what the `return` will be doing.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+                f'Found {possible_hits[-1]} as restart file for iteration {iteration} and walker {walker} based on mdtraj extensions. if this is incorrect, provide a file name using flag --rst-file'
+            )
+        if len(possible_hits) > 1:
+            possible_hits = sorted(possible_hits, key=lambda file: os.path.getctime(os.path.join(search_folder, file)))
+            log.warning(
+                f'Found {possible_hits[-1]} as restart file for iteration {iteration} and walker {walker} based on file creation times. if this is incorrect, provide a file name using flag --rst-file'
+            )
+        rst_dest_name = f"{iteration:06d}_{walker:06d}.{possible_hits[-1].split('.')[-1]}"
+        shutil.copyfile(os.path.join(search_folder, possible_hits[-1]), os.path.join(self.output_bstates_dir, rst_dest_name))
+        return rst_dest_name
```
jeremyleung521, line 221: L198 - L230

I think it's better to structure this so it's three if/elif/else blocks to better imply that they are three different methods for determining restart files of different priority. Right now it's not very clear/readable, especially without line breaks and reusing of variable names etc.

Might even be better to just leave one copy of the return `rst_dest_name` at the bottom, since it's consistently the same return line!

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        search_folder: path or string describing a path to the directory that will be searched for a trajectory file
+
+        iteration: integer describing the iteration number of the trajectory
+
+        walker: integer describing the segment id of the trajectory
+
+        Returns
+        -------
+        rst_dest_name: string describing the name of the copied trajectory in the output_bstates_dir
+        '''
```
jeremyleung521:
```suggestion
        
        """
```

While not consistent in the WESTPA code, leave a blank line after docstring (before the triple quotes)

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+                    index,
+                    self.h5['iterations'][iteration]['seg_index']['weight'][index],
+                ]
+                for index in indices
+            ]
+            succ += temp_array
+        return np.asarray(succ)
+
+    def copy_traj(self, search_folder, iteration, walker):
+        '''
```
jeremyleung521:
```suggestion
        """
```
I know it's not very consistent in the WESTPA codebase, but triple quotes for Numpy docstrings.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+                f'Found {possible_hits[-1]} as restart file for iteration {iteration} and walker {walker} based on file creation times. if this is incorrect, provide a file name using flag --rst-file'
+            )
+        rst_dest_name = f"{iteration:06d}_{walker:06d}.{possible_hits[-1].split('.')[-1]}"
+        shutil.copyfile(os.path.join(search_folder, possible_hits[-1]), os.path.join(self.output_bstates_dir, rst_dest_name))
+        return rst_dest_name
+
+    def go(self):
+        """
+        Main public method for running w_reverse. Runs w_succ to find the successful trajectories then iterates over them to copy the trajectories to the output_bstates_dir. Then it iterates over the trajectories again to make the output_bstates_file.
+        """
```
jeremyleung521:
```suggestion
        '''Main public method for running w_reverse. Runs w_succ to find the successful trajectories then iterates over them to copy the trajectories to the output_bstates_dir. Then it iterates over the trajectories again to make the output_bstates_file.
        '''
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+                files = [file for file in os.listdir(self.output_bstates_dir) if file.startswith(f'{iteration:06d}_{walker:06d}')]
+                rst_dest_name = ''
+                if len(files) > 1:
+                    if f'{iteration:06d}_{walker:06d}.{self.rst_extension}' in files:
+                        rst_dest_name = f'{iteration:06d}_{walker:06d}.{self.rst_extension}'
+                    else:
+                        files = sorted(files, key=lambda file: os.path.getctime(file))
+                        rst_dest_name = files[-1]
+                        log.warning(
+                            f'Muliple files starting with {iteration:06d}_{walker:06d} are in the output directory. Using {rst_dest_name}. if this is incorrect, provide a file name using flag --rst-file so that the correct extension can be used'
```
jeremyleung521:
```suggestion
                            f'Muliple files starting with {iteration:06d}_{walker:06d} are in the output directory. Using {rst_dest_name=}. if this is incorrect, provide a file name using flag `--rst-file` so that the correct extension can be used'
```

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        # Number of reverse bstates created
+        # different totals if the max is less than total succ_pairs to loop
+        total_pairs = min(self.max_n_bstates, len(succ_pairs))
+        # then for each pair
+        rng = np.random.default_rng(self.seed)
+        succ_pairs_used = rng.choice(
+            succ_pairs, size=total_pairs, p=succ_pairs[:, 2] / np.sum(succ_pairs[:, 2], dtype=float), replace=False
+        )
+        total_weight = 0
+        for idx, succ_pair_used in tqdm(enumerate(succ_pairs_used), total=total_pairs, desc="New bstates"):
```
jeremyleung521:
```suggestion
        for idx, succ_pair_used in enumerate(tqdm(succ_pairs_used, total=total_pairs, desc="New bstates")):
```
enumerate outside tqdm is the suggested form

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,302 @@
...
+        os.makedirs(self.output_bstates_dir, exist_ok=True)
+        # Number of reverse bstates created
+        # different totals if the max is less than total succ_pairs to loop
+        total_pairs = min(self.max_n_bstates, len(succ_pairs))
+        # then for each pair
+        rng = np.random.default_rng(self.seed)
+        succ_pairs_used = rng.choice(
+            succ_pairs, size=total_pairs, p=succ_pairs[:, 2] / np.sum(succ_pairs[:, 2], dtype=float), replace=False
+        )
+        total_weight = 0
```
jeremyleung521:
```suggestion
        total_weight = 0.
```

Might as well initiate it as a float, since we know it'll be a float.

setup.py
```diff
@@ -88,6 +88,7 @@ def extensions():
     'w_pdist = westpa.cli.tools.w_pdist:entry_point',
     'w_ipa = westpa.cli.tools.w_ipa:entry_point',
     'w_bins = westpa.cli.tools.w_bins:entry_point',
+    'w_reverse = westpa.cli.tools.w_reverse:entry_point',
```
jeremyleung521: place it at the bottom. It's easier that way to determine _when_ it's been developed.

tests/test_tools/test_w_reverse.py
```diff
@@ -0,0 +1,127 @@
...
+            ),
+        ):
+            entry_point()
+        assert os.path.isfile('./bstates_reverse/bstates.txt'), "The bstates.txt file was not generated."
+        assert os.path.isfile('./bstates_reverse/000001_000000.xml'), "The 000001_000000.xml file was not generated."
+        assert os.path.getsize('./bstates_reverse/000001_000000.xml') > 0, "The 000001_000000.xml file is empty."
+        assert os.path.isfile('./bstates_reverse/000002_000000.xml'), "The 000002_000000.xml file was not generated."
+        assert os.path.getsize('./bstates_reverse/000002_000000.xml') > 0, "The 000002_000000.xml file is empty."
+        assert os.path.isfile('./bstates_reverse/000003_000000.xml'), "The 000003_000000.xml file was not generated."
+        assert os.path.getsize('./bstates_reverse/000003_000000.xml') > 0, "The 000003_000000.xml file is empty."
```
jeremyleung521:
```suggestion
        for iiter in range(1,4):
            assert os.path.isfile(f'./bstates_reverse/{iiter:06d}_000000.xml'), "The {iiter:06d}_000000.xml file was not generated."
            assert os.path.getsize(f'./bstates_reverse/{iiter:06d}_000000.xml') > 0, f"The {iiter:06d}_000000.xml file is empty."
```

tests/test_tools/test_w_reverse.py
```diff
@@ -0,0 +1,127 @@
...
+        assert cmp(
+            os.path.join(ref_dir, 'bstates.txt'), './bstates_reverse/bstates.txt'
+        ), 'The reference bstates.txt and the produced bstates.txt are not the same'
+        shutil.rmtree('traj_segs')
+        os.remove('west.h5')
+        os.remove('west.cfg')
```
jeremyleung521, line 126: Tests are not running in a tmp directory. You might be deleting/overwriting files.

tests/refs/west_reverse_no_hdf5.cfg
jeremyleung521, line 1: These two cfg files have like one line difference. Do something like this to strip out the line.

https://github.com/westpa/westpa/blob/westpa2/tests/conftest.py#L225-L228

jeremyleung521 (changes requested):
The tests need more work (we don't use bare unittest anymore and have migrated towards pytest, so do that) and need to use temp directories. I would also have separate tests for each method, especially the file parsing parts.

The rest are mostly stylistic choices (most you could probably just take the suggestion). `copy_traj` in particular is not very readable and require some refactoring.

Will actually test this out with files in a bit.

src/westpa/cli/tools/w_reverse.py
```diff
@@ -0,0 +1,288 @@
...
+import os
+import shutil
+import tempfile
+from westpa.core.h5io import WESTIterationFile, WESTPAH5File
+from westpa.core.propagators.loaders import restart_writer
+from westpa.core.segment import Segment
+from westpa.core._rc import WESTRC
+from westpa.tools import WESTTool, WESTDataReader
+import numpy as np
+from westpa.core.trajectory import mdtraj_supported_extensions
```
jeremyleung521:
```suggestion
import logging
import os
import shutil
import tempfile

import numpy as np
from tqdm.auto import tqdm

from westpa.core._rc import WESTRC
from westpa.core.h5io import WESTIterationFile, WESTPAH5File
from westpa.core.propagators.loaders import restart_writer
from westpa.core.segment import Segment
from westpa.core.trajectory import mdtraj_supported_extensions
from westpa.tools import WESTTool, WESTDataReader
```

Alphabetical order + separated into three sections: "standard libraries, other dependencies, local/self imports" per PEP 8
https://peps.python.org/pep-0008/#imports

jeremyleung521 (approved):

